PR #10, opened 2019-12-27
Maxchehab/ch1169/remove redirect uri from the profile function

maxchehab:
This PR:
- removes the `redirect_uri` parameter from the `WorkOS::SSO.profile` function and removes the `redirect_uri` query parameter from the `POST /sso/token` request.
- adds a [**Release Notes**](https://github.com/workos-inc/workos-ruby/tree/maxchehab/ch1169/remove-redirect-uri-from-the-getprofile-function#release-notes) section to the readme with migration examples.


README.md
```diff
@@ -19,3 +20,28 @@ gem build workos.gemspec
...
 - Ruby 2.3+.
+
+# Release Notes
+
+### December 27, 2019
+
+- Removed the `redirect_uri` parameter from the `WorkOS::SSO.profile` function. Migrating existing code just requires removing the existing parameter:
+
+  ```ruby
+  # < December 27, 2019
```
maxchehab: Not sure how versioning works w/ Ruby. I'm sure there is a more idiomatic way, but not sure where the current version is located in the repo.

sbauch: couple of moving parts with the versioning

version is defined here: https://github.com/workos-inc/workos-ruby/blob/master/lib/workos/version.rb

I think technically this would be a breaking minor change, so we'd go to `v0.1.0` with this?

but since we have folks installing the gem via github, we need to wait to merge this until we can get in touch with our customers.

They could update to 0.0.2 from ruby gems (releasing that now) or we could have them implement these changes and continue installing from github or push a new version to rubygems. a little messy since we are touch and go with a first customer. I'm gonna link them to this PR

maxchehab: Sweet. I'll go ahead and bump the version in that file and update the readme. We can hold off on releasing this version till everything is ready to go.

sbauch: we could maybe also start using github releases for changelog?

alos, you'll want to update yard docs here, theres a bash script in bin/docs that reads the comment style documentation and creates the files for https://workos-inc.github.io/workos-ruby/

sbauch:
we're unblocked here and can start merging things for our v0.1.0 tag / release.

for this PR, we need to:

1. update github pages documentation - just run `bin/docs` and commit the result
2. implement what we decided wrt tags / releases / changelog. I think that means removing changelog things from README here, and setting up releases in the repo. is that where we landed?

maxchehab:
Just updated the documentation. Also added `yard` as a development dependency.

> Removing changelog things from README here, and setting up releases in the repo. is that where we landed?

Yep, that's the plan. [Just created a pre-release here](https://github.com/workos-inc/workos-ruby/releases/tag/0.1.0).

PR #13, opened 2019-12-30
Attach the Request ID to relevant SDK exceptions

sbauch:
https://app.clubhouse.io/workos/story/1178/ruby-attach-the-request-id-to-relevant-sdk-exceptions

---

Pretty simple! Couple of lines + a spec.

`Net::HTTP` is a little odd in the sense that it downcases headers. All good though:

```ruby
irb(main):015:0>   profile = WorkOS::SSO.profile(
irb(main):016:1*     code: 'foo',
irb(main):017:1*     project_id: PROJECT_ID,
irb(main):018:1*     redirect_uri: REDIRECT_URI
irb(main):019:1>   )
...
WorkOS::RequestError (The code 'foo' has expired or is invalid. - request ID: 7423a232-c8d7-41b5-bb2e-375d6919a019)
```

Totally open to suggestions wrt to the formatting or copy of adding the request ID to the error message. Ruby exceptions just take a string message, nothing formatted like key / value.

@mark I figure this `client` method will become a singleton class in order to support Audit Logs fwiw.

### DIFF
--- a/spec/lib/workos/sso_spec.rb
+++ b/spec/lib/workos/sso_spec.rb
@@ -83,7 +83,7 @@ describe WorkOS::SSO do
         stub_request(:post, 'https://api.workos.com/sso/token').
           with(query: query).
           to_return(
-            headers: { 'X-Request-ID' => 'heroku-request-id' },
+            headers: { 'X-Request-ID' => 'request-id' },
             status: 422,
             body: { "message": 'some error message' }.to_json,
           )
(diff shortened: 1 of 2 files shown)


spec/lib/workos/sso_spec.rb
```diff
@@ -83,23 +83,30 @@
         stub_request(:post, 'https://api.workos.com/sso/token').
           with(query: query).
           to_return(
+            headers: { 'X-Request-ID' => 'heroku-request-id' },
```
marktran:
```suggestion
            headers: { 'X-Request-ID' => 'request-id' },
```

Nit: `X-Request-ID` is currently a uuid generated by Heroku but if we moved to another platform, it'd be generated by something else. Maybe something more generic instead?

sbauch:
Good call, will fix.

🚨 We also can't merge this yet (until humbledot is off installing via github)

PR #14, opened 2020-01-02
Add version header to Ruby SDK requests

sbauch:
https://app.clubhouse.io/workos/story/1162/add-ruby-sdk-version-to-user-agent-header

---

Adds the `'X-SDK-VERSION'` header to outgoing requests (just profile for now with SSO).

Is that the best name for the header?

EDIT:

This now puts the version and some other info (ruby version, ruby platform) in the User Agent:
```
'User-Agent'=>'WorkOS; ruby/2.6.5; x86_64-darwin18; v0.0.2'
```

I cribbed the method from another gem that seemed legit!

I've also removed codecov "patch" checking and added the badge to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # workos-rb
 WorkOS official Ruby gem for interacting with WorkOS APIs
 
+[![codecov](https://codecov.io/gh/workos-inc/workos-ruby/branch/master/graph/badge.svg)](https://codecov.io/gh/workos-inc/workos-ruby)
+
 ## Installation
 
 You don't need this source code unless you want to modify the gem. If you just
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
-# workos-rb
-WorkOS official Ruby gem for interacting with WorkOS APIs
+# workos-ruby [![codecov](https://codecov.io/gh/workos-inc/workos-ruby/branch/master/graph/badge.svg)](https://codecov.io/gh/workos-inc/workos-ruby)
 
-[![codecov](https://codecov.io/gh/workos-inc/workos-ruby/branch/master/graph/badge.svg)](https://codecov.io/gh/workos-inc/workos-ruby)
+WorkOS official Ruby gem for interacting with WorkOS APIs
 
 ## Installation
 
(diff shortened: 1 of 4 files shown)


sbauch:
[ch1162]

lib/workos/sso.rb
```diff
@@ -124,6 +124,13 @@ def client
         @client
       end
 
+      sig { params(path: String).returns(Net::HTTP::Post) }
+      def post_request(path:)
+        request = Net::HTTP::Post.new(path)
+        request['X-SDK-VERSION'] = WorkOS::VERSION
```
marktran: What do you think about just stuffing everything into the user agent? This is what Stripe/browser/HTTP-clients do

https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/stripe_client.rb#L852

sbauch: @maxchehab wdyt about this? i think the goal is to make things more discoverable in logs - as long as the user agent string is searchable in logs I wouldn't mind it

maxchehab: =I think we should just use the user agent header here. It will make it a little easier for filtering in the future if all versions are stored in the same header, including requests from the browser etc.

sbauch: kk back to the drawing board! 📝

README.md
```diff
@@ -1,6 +1,8 @@
 # workos-rb
```
marktran:
```suggestion
# workos-ruby
```

README.md
```diff
@@ -1,6 +1,8 @@
...
 WorkOS official Ruby gem for interacting with WorkOS APIs
 
+[![codecov](https://codecov.io/gh/workos-inc/workos-ruby/branch/master/graph/badge.svg)](https://codecov.io/gh/workos-inc/workos-ruby)
```
marktran: How about putting the badget next to the main header?

https://github.com/workos-inc/workos/raw/master/README.md

lib/workos/sso.rb
```diff
@@ -124,6 +124,25 @@ def client
         @client
       end
 
+      sig { params(path: String).returns(Net::HTTP::Post) }
+      def post_request(path:)
```
marktran: Going to extract all of this into a common client setup file for use with the audit log module

PR #18, opened 2020-01-29
add provider param, making it and domain optional

sbauch:
Changes to support "Sign in With Google" buttons where the Developer does not know the domain, but does know the provider.

Idk if I got too fancy with the Provider enum stuff. Had hoped it would work a little nicer where we could define an Enum, and then use it in the method signature, but pass the method strings. Doesn't quite work that way unfortunately.

Will build docs after approval i guess? lots of noise for review


lib/workos/types/provider_enum.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+# typed: strict
+
+module WorkOS
+  module Types
+    # The ProviderEnum is type-safe declarations of a
+    # fixed set of values for SSO Providers.
+    class Provider < T::Enum
+      enums do
+        Google = new('Google')
```
sbauch: i took this from the node PR - should it be `GoogleOAuth`?

marktran: Yeah, I think `GoogleOAuth` makes sense

PR #19, opened 2020-02-11
Add DraftConnection support

sbauch:
v low priority here. mostly wanted to get your 👀 here for finding a better naming approach. I think that until we move further forward with IdP Link embed we'll not cut a new gem version, but this is for the ability to demo.

### DIFF
--- a/lib/workos/sso.rb
+++ b/lib/workos/sso.rb
@@ -140,8 +140,7 @@ module WorkOS
       end
       def promote_draft_connection(token:)
         request = bearer_post_request(
-          path: '/draft_connections/convert',
-          body: { id: token },
+          path: "/draft_connections/#{token}/convert",
         )
 
         response = client.request(request)
--- a/lib/workos/sso.rb
+++ b/lib/workos/sso.rb
@@ -196,7 +196,7 @@ module WorkOS
 
       def bearer_post_request(path:, body: nil)
         request = post_request(path: path, body: body)
-        request['Authorization'] = "Bearer #{WorkOS.key}"
+        request['Authorization'] = "Bearer #{WorkOS.key!}"
         request
       end
     end
--- a/lib/workos/sso.rb
+++ b/lib/workos/sso.rb
@@ -194,7 +194,7 @@ module WorkOS
       end
       # rubocop:enable Metrics/MethodLength
 
-      def bearer_post_request(path:, idempotency_key: nil, body: nil)
+      def bearer_post_request(path:, body: nil)
         request = post_request(path: path, body: body)
         request['Authorization'] = "Bearer #{WorkOS.key}"
         request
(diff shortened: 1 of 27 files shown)


lib/workos/sso.rb
```diff
@@ -119,6 +119,36 @@ def profile(code:, project_id:)
...
+      # @return [Bool] - returns `true` if successful, `false` otherwise.
+      # @see https://github.com/workos-inc/idp-link
+      sig do
+        params(
+          token: String,
+        ).returns(T::Boolean)
+      end
+      def promote_draft_connection(token:)
+        request = bearer_post_request(
+          path: '/draft_connections/convert',
```
marktran: https://github.com/workos-inc/workos/pull/2288

Changed it to `/draft_connections/activate`

spec/lib/workos/sso_spec.rb
```diff
@@ -210,4 +210,45 @@
...
+    let(:body) do
+      {
+        id: token,
+      }
+    end
+
+
+    context 'with a valid request' do
+      before do
+        stub_request(:post, 'https://api.workos.com/draft_connections/convert').
```
marktran: activate

spec/lib/workos/sso_spec.rb
```diff
@@ -210,4 +210,45 @@
...
+          token: token,
+        )
+
+        expect(response).to be(true)
+      end
+    end
+
+    context 'with an invalid request' do
+      before do
+        stub_request(:post, 'https://api.workos.com/draft_connections/convert').
```
marktran: activate

lib/workos/sso.rb
```diff
@@ -119,6 +119,35 @@ def profile(code:, project_id:)
...
+      #   => true
+      #
+      # @return [Bool] - returns `true` if successful, `false` otherwise.
+      # @see https://github.com/workos-inc/idp-link
+      sig do
+        params(
+          token: String,
+        ).returns(T::Boolean)
+      end
+      def promote_draft_connection(token:)
```
marktran: Can we update this to `activate_draft_connection`?

And something other than `token` since it's actually the resource id. `draft_connection_id`?

spec/lib/workos/sso_spec.rb
```diff
@@ -210,4 +210,37 @@
       end
     end
   end
+
+  describe '.promote_draft_connection' do
+    let(:token) { 'draft_conn_942u509ut' }
```
marktran: `draft_connection_id`?

lib/workos/sso.rb
sbauch: that breaks so much documentation etc. we've always said we were gonna call it a token

spec/lib/workos/sso_spec.rb
sbauch:
```suggestion
    let(:token) { 'draft_conn_id' }
```

PR #21, opened 2020-03-13
Update comments referencing IdP Link to WorkOS.js

henrylamchan:
# Overview
This PR updates references to IdP Link to WorkOS.js


sbauch:
can you run `bin/docs`? that updates the generated html (will also update some timestamps so that its clear the docs are up to date)

PR #31, opened 2020-04-24
add WorkOS::SSO.create_connection; refactor

marktran:


lib/workos/connection.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+# typed: true
+
+require 'json'
+
+module WorkOS
+  # The Connection class provides a lightweight wrapper around
+  # a WorkOS Connection resource. This class is not meant to be instantiated
+  # in user space, and is instantiated inernally but exposed.
```
henrylamchan, line 9:
```suggestion
  # in user space, and is instantiated internally but exposed.
```

lib/workos/sso.rb
```diff
@@ -168,6 +188,20 @@ def validate_domain_and_provider(domain:, provider:)
           " `provider` must be in #{PROVIDERS}"
       end
 
+      sig { params(response: Net::HTTPResponse).void }
+      def check_and_raise_error(response:)
```
henrylamchan, line 192: Noticed the client doesn't handle a 404. Would it helpful for this logic to be moved down there?

lib/workos/types/connection_struct.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+# typed: strict
+
+module WorkOS
+  module Types
+    # This ConnectoinStruct acts as a typed interface
```
henrylamchan, line 6:
```suggestion
    # This ConnectionStruct acts as a typed interface
```

lib/workos/sso.rb
marktran: Do you mean to where the request is executed so it covers all requests?

henrylamchan: Yep. Can't really think of any other cases where we might get a 404 but the client seems like a good home for that logic. Would getting a user potentially result in a 404 too if bad values are passed in?

marktran: 👍 good idea, I'll move it there

We'll return a 404 if the source doesn't exist or there's an authorization failure

PR #38, opened 2020-05-20
change type

marktran:
This should be a string.


rjvani:
Do you need to update parameters: https://github.com/workos-inc/workos-ruby/blob/d910c4c09e318f66daed132eadc90e572ebc885d/lib/workos/sso.rb#L62?

PR #43, opened 2020-07-31
Support for GET /events endpoint

willmanduffy:
* Adding the ability to fetch Audit Trail events via the Ruby SDK.
* Bump gem version to 0.4.0

### DIFF
--- a/spec/support/fixtures/vcr_cassettes/audit_trail/get_events.yml
+++ b/spec/support/fixtures/vcr_cassettes/audit_trail/get_events.yml
@@ -14,7 +14,7 @@ http_interactions:
         Accept:
           - "*/*"
         User-Agent:
-          - WorkOS; ruby/2.7.1; x86_64-darwin19; v0.3.3
+          - WorkOS; ruby/2.7.1; x86_64-darwin19; v0.4.0
         Authorization:
           - Bearer <API_KEY>
     response:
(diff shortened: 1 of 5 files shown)


spec/support/fixtures/vcr_cassettes/audit_trail/get_events.yml
```diff
@@ -0,0 +1,61 @@
...
+        string: ""
+      headers:
+        Content-Type:
+          - application/json
+        Accept-Encoding:
+          - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+        Accept:
+          - "*/*"
+        User-Agent:
+          - WorkOS; ruby/2.7.1; x86_64-darwin19; v0.3.3
```
marktran, line 17: You'll need to bump the version here as well—this will fail somewhere down the line

willmanduffy: Wild! Curious to hear which cases they would fail I thought the user agent in vcr was largely incidental. I've gone ahead and done so. I notice all of the other cassettes have older versions too do we need to update them as well?

marktran: Oh maybe it doesn't 🤔 

I recall versioning causing failures but I'm not sure where

PR #46, opened 2020-08-24
Expose raw_attributes on WorkOS::Profile

willmanduffy:
* On the API, we recently exposed raw_attributes, the original profile
response we receive from the IdP. This update passes this data through
the SDK.
* Updates the SDK to 0.5.0


willmanduffy:
@henrylamchan Yup! Just rebased off of it!

PR #48, opened 2020-09-09
Add WorkOS::Portal.list_organizations

willmanduffy:
* Returns a list of the organizations accessible to the provided
  API key. Includes pagination functionality via the list_metadata
  property.

### DIFF
--- a/lib/workos/types.rb
+++ b/lib/workos/types.rb
@@ -6,6 +6,7 @@ module WorkOS
   # so we're using Sorbet throughout this Ruby gem.
   module Types
     require_relative 'types/connection_struct'
+    require_relative 'types/list_struct'
     require_relative 'types/profile_struct'
     require_relative 'types/provider_enum'
   end
--- a/lib/workos.rb
+++ b/lib/workos.rb
@@ -31,6 +31,7 @@ module WorkOS
   autoload :AuditTrail, 'workos/audit_trail'
   autoload :Connection, 'workos/connection'
   autoload :DirectorySync, 'workos/directory_sync'
+  autoload :Portal, 'workos/portal'
   autoload :Profile, 'workos/profile'
   autoload :SSO, 'workos/sso'
 
--- a/lib/workos/portal.rb
+++ b/lib/workos/portal.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+# typed: true
+
+require 'net/http'
+
+module WorkOS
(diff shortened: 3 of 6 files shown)


lib/workos/organization.rb
```diff
@@ -0,0 +1,51 @@
...
+      #  organizations after the provided Organization ID.
+      # @param [Integer] limit A pagination argument used to limit the number
+      #  of listed Organizations that are returned.
+      sig do
+        params(
+          options: T::Hash[Symbol, String],
+        ).returns(WorkOS::Types::ListStruct)
+      end
+      # rubocop:disable Metrics/MethodLength
+      def list(options = {})
```
willmanduffy, line 32: I know we didn't land on a name so I dropped this in temporarily. Happy to adjust to the find decided name!

:yak_shaving:

lib/workos/types.rb
```diff
@@ -6,6 +6,7 @@ module WorkOS
   # so we're using Sorbet throughout this Ruby gem.
   module Types
     require_relative 'types/connection_struct'
+    require_relative 'types/list_struct'
```
willmanduffy, line 9: This really threw me though a loop and lost a couple hours to finding this...

Thoughts on pulling `require_all`?

https://github.com/jarmo/require_all

spec/lib/workos/organization_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+          '/organizations?before=before-id',
+          'Content-Type' => 'application/json'
+        ]
+
+        expected_request = Net::HTTP::Get.new(*request_args)
+
+        expect(Net::HTTP::Get).to receive(:new).with(*request_args).
+          and_return(expected_request)
+
+        VCR.use_cassette 'organization/list', match_requests_on: [:path] do
```
willmanduffy, line 42: I decided to use the same cassette for all the tests since the response is handled the same regardless. What we really want to test is the request we are sending out and trusting the API will Do The Right Thing™. `expected_request` is the real test in these option tests.

willmanduffy:
@marktran @maxchehab This has been updated with the new naming we outlined in our call today.

marktran (approved):

lib/workos/types.rb
willmanduffy: Will follow up with this in a future PR to avoid merge conflicts.

PR #53, opened 2020-09-14
Use require_all to pull in types folder

willmanduffy:
* Resolves a hidden step needed before adding a file to the types
  folder that can be confusing for developers new to the SDK
  codebase.
* Discussion here: https://github.com/workos-inc/workos-ruby/pull/48#discussion_r485866330

### DIFF
--- a/lib/workos/types.rb
+++ b/lib/workos/types.rb
@@ -1,16 +1,11 @@
 # frozen_string_literal: true
-# typed: strong
+
+require 'require_all'
 
 module WorkOS
   # WorkOS believes strongly in typed languages,
   # so we're using Sorbet throughout this Ruby gem.
   module Types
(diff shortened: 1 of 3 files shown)


lib/workos/types.rb
```diff
@@ -1,15 +1,11 @@
 # frozen_string_literal: true
-# typed: strong
```
willmanduffy, line 2: Had to remove this from this file to get `srb tc` to pass on `require_all`. Would the real solution be to write types for it?

https://sorbet.org/docs/error-reference#7018

marktran: 🤔 Not sure, I don't know Sorbet that well tbh

marktran (approved):

PR #55, opened 2020-09-16
Add MagicLink support through Passwordless module

zlwaterfield:
Add Passwordless support for MagicLink. Adds two methods
 - `create_session` - generate a passwordless session for a user from their email
 - `send_session` - send an email to a user with a Magic Link confirmation URL

### DIFF
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -109,8 +109,8 @@ module WorkOS
         )
       when 422
         errors = nil
-        if json['errors']
-          errors = json['errors'].map do |error|
+        errors = if json['errors']
+          json['errors'].map do |error|
             "#{error['field']}: #{error['code']}"
           end.join('; ')
         end
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -81,12 +81,6 @@ module WorkOS
       ].join('; ')
     end
 
-    def extract_error(errors)
-      errors.map do |error|
-        "#{error['field']}: #{error['code']}"
-      end.join('; ')
-    end
-
     # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Metrics/CyclomaticComplexity
     sig { params(response: ::T.untyped).void }
     def handle_error_response(response:)
@@ -125,5 +119,13 @@ module WorkOS
       end
     end
     # rubocop:enable Metrics/MethodLength, Metrics/AbcSize, Metrics/CyclomaticComplexity
+
+    private
+
+    def extract_error(errors)
+      errors.map do |error|
+        "#{error['field']}: #{error['code']}"
+      end.join('; ')
+    end
   end
 end
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -115,7 +115,7 @@ module WorkOS
           end.join('; ')
         end
 
-        message = "#{json['message']}"
+        message = json['message']
         message += " (#{errors})" if errors
         raise InvalidRequestError.new(
           message: message,
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -108,11 +108,15 @@ module WorkOS
           request_id: response['x-request-id'],
         )
       when 422
-        errors = json['errors'].map do |error|
-          "#{error['field']}: #{error['code']}"
-        end.join('; ')
+        errors = nil
+        if json['errors']
+          errors = json['errors'].map do |error|
+            "#{error['field']}: #{error['code']}"
+          end.join('; ')
+        end
 
-        message = "#{json['message']} (#{errors})"
+        message = "#{json['message']}"
+        message += " (#{errors})" if errors
         raise InvalidRequestError.new(
           message: message,
           http_status: http_status,
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -81,8 +81,13 @@ module WorkOS
       ].join('; ')
     end
 
+    def extract_error(errors)
+      errors.map do |error|
+        "#{error['field']}: #{error['code']}"
+      end.join('; ')
+    end
 
-    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Metrics/CyclomaticComplexity
     sig { params(response: ::T.untyped).void }
     def handle_error_response(response:)
       http_status = response.code.to_i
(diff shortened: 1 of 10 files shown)


lib/workos/client.rb
```diff
@@ -108,11 +108,15 @@ def handle_error_response(response:)
           request_id: response['x-request-id'],
         )
       when 422
-        errors = json['errors'].map do |error|
-          "#{error['field']}: #{error['code']}"
-        end.join('; ')
-
-        message = "#{json['message']} (#{errors})"
+        errors = nil
```
willmanduffy, line 111: Curious to hear context on what you encountered to need this change!

Also, you should be able to do

```rb
errors = if json['errors']
  json['errors'].map do |error|
    "#{error['field']}: #{error['code']}"
  end.join('; ')
end
```

This will put `nil` in `errors` if `json['errors']` is not present.

zlwaterfield:
@marktran any idea why this is too complex and how to fix it? https://workos.semaphoreci.com/jobs/0fa8dea1-6231-4767-b19e-f1af61f801ae (https://github.com/workos-inc/workos-ruby/pull/55/files#diff-263cb07c6a852fd6a6a5e9053a0ce8ecR111)

lib/workos/client.rb
```diff
@@ -108,11 +108,15 @@ def handle_error_response(response:)
...
+        if json['errors']
+          errors = json['errors'].map do |error|
+            "#{error['field']}: #{error['code']}"
+          end.join('; ')
+        end
+
+        message = json['message']
+        message += " (#{errors})" if errors
```
willmanduffy, line 119: This is perfectly fine as is. If you're feeling fancy you could do 

```rb
[message, errors].compact.join(' ')
```

Example:

```
irb(main):001:0> foo = "bar"
=> "bar"
irb(main):002:0> baz = nil
=> nil
irb(main):003:0> [foo, baz].compact.join(' ')
=> "bar"
irb(main):004:0> baz = "bits"
=> "bits"
irb(main):005:0> [foo, baz].compact.join(' ')
=> "bar bits"
```

compact will remove nil items from the array.

lib/workos/passwordless.rb
```diff
@@ -0,0 +1,71 @@
...
+      #  encode arbitrary information to help restore application state between
+      #  redirects.
+      # @option options [String] type The type of Passwordless Session to
+      #  create. Currently, the only supported value is 'MagicLink'.
+      #
+      # @return Hash
+      sig do
+        params(
+          options: Hash,
+        ).returns(::T.untyped)
```
willmanduffy, line 34: I wonder if we want to make a `PasswordlessSessionStruct` and return that. See an example in `OrganizationStruct`

lib/workos/passwordless.rb
```diff
@@ -0,0 +1,71 @@
...
+  # API key.
+  #
+  # @see https://workos.com/docs/sso/configuring-magic-link
+  module Passwordless
+    class << self
+      extend T::Sig
+      include Base
+      include Client
+
+      # Create an Passwordless Session.
```
willmanduffy, line 19:
```suggestion
      # Create a Passwordless Session.
```

lib/workos/client.rb
zlwaterfield: The `send_session` error response only have a `message` key and no `errors` key so `json['errors'].map` was throwing an error. I was attempting to only run the map if `json['errors']` is present.

lib/workos/client.rb
zlwaterfield: Nice, thanks!

lib/workos/passwordless.rb
zlwaterfield: - [ ] Add `PasswordlessSessionStruct`

lib/workos/passwordless.rb
zlwaterfield: - [ ] Fix grammar mistake

lib/workos/client.rb
zlwaterfield: - [ ] Update message to use `[message, errors].compact.join(' ')`

lib/workos/passwordless.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+# typed: true
+
+require 'net/http'
+require 'uri'
```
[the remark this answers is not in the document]
zlwaterfield: - [ ] remove 'uri'

lib/workos/passwordless.rb
```diff
@@ -0,0 +1,82 @@
...
+          request: post_request(
+            path: '/passwordless/sessions',
+            auth: true,
+            body: options,
+          ),
+        )
+
+        hash = JSON.parse(response.body)
+
+        WorkOS::Types::PasswordlessSessionStruct.new(
```
zlwaterfield, line 48: @willmanduffy, curious your thoughts on this?

lib/workos/client.rb
```diff
@@ -108,18 +113,17 @@ def handle_error_response(response:)
           request_id: response['x-request-id'],
         )
       when 422
-        errors = json['errors'].map do |error|
-          "#{error['field']}: #{error['code']}"
-        end.join('; ')
+        message = json['message']
+        errors = extract_error(json['errors']) if json['errors']
```
zlwaterfield, line 117: @willmanduffy this is how I ended up getting it to work

lib/workos/types/passwordless_session_struct.rb
```diff
@@ -0,0 +1,16 @@
...
+module WorkOS
+  module Types
+    # This PasswordlessSessionStruct acts as a typed interface
+    # for the Passwordless class
+    class PasswordlessSessionStruct < T::Struct
+      const :id, String
+      const :email, String
+      const :expires_at, Date
+      const :link, String
+      const :object, String
```
willmanduffy, line 13: I think we can leave off `object` this is metadata from the API we don't need to expose

lib/workos/client.rb
```diff
@@ -81,8 +81,13 @@ def user_agent
       ].join('; ')
     end
 
+    def extract_error(errors)
```
willmanduffy, line 84: Let's throw this at the bottom of the file as a private method.

```rb
def public_method
end

private

def extract_error(errors)
end
```

zlwaterfield: - [ ] move method to bottom and make private

lib/workos/types/passwordless_session_struct.rb
zlwaterfield: - [ ] remove object metadata field

willmanduffy (approved):

PR #60, opened 2020-10-13
Expose `connection_id` on Profile Objects

jthodge:
This PR introduces the following changes:

* Passes the `connection_id` Profile object attribute through to the SDK. We currently expose `connection_id` via the `/sso/token` API endpoint.
* Updates the SDK version to `0.9.0`.


lib/workos/version.rb
```diff
@@ -2,5 +2,5 @@
 # typed: strong
 
 module WorkOS
-  VERSION = '0.8.1'
+  VERSION = '0.9.0'
```
willmanduffy, line 5: I think we should be good for this to be 0.8.2 as it's not a breaking change of any kind

willmanduffy (approved):

lib/workos/version.rb
jthodge: @willmanduffy Hm I do think this would be minor bump instead of a patch based on our [versioning policy](https://github.com/workos-inc/workos/wiki/SDK-Release-Process#versioning-policy). Since you're right, it doesn't include breaking changes, and also isn’t a bug fix 🤔

willmanduffy: Yeah, I think that's totally right if we're past 1.0.0. I personally think that versioning policy is a bit aggressive for the amount of change we're doing right now. Semver points out that pre-1.0.0 can be the wild west

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

https://semver.org/#spec-item-4

That said if all the others are merged as minor bumps right now we should go ahead with that for consistency

jthodge: Ah, got it! Thank you, I appreciate the spec pointer.
I'm on board with your proposal: we can merge this one as a minor bump for consistency. (And I'll be more conservative with versioning moving forward as well.)

lib/workos/profile.rb
```diff
@@ -13,9 +13,9 @@ class Profile
     extend T::Sig
 
     sig { returns(String) }
-    attr_accessor :id, :email, :first_name, :last_name,
+    attr_accessor :id, :email, :first_name, :last_name, :connection_id,
                   :connection_type, :idp_id, :raw_attributes
-
+    # rubocop:disable Metrics/AbcSize
```
jthodge, line 18: @willmanduffy @marktran How do we feel about this rubocop configuration change instead of pulling assignments out into `before_action`?

lib/workos/profile.rb
```diff
@@ -48,7 +51,7 @@ def to_json(*)
 
     private
 
-    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
jthodge, line 54: @willmanduffy @marktran How do we feel about this rubocop configuration change since it's a pretty large hash?

willmanduffy: These are both good! This cop is a bit aggressive to be honest!

lib/workos/profile.rb
jthodge: https://github.com/workos-inc/workos-ruby/pull/60#discussion_r505659519

lib/workos/profile.rb
jthodge: sgtm! 👍

PR #64, opened 2020-12-09
Authenticate POST /events endpoint

croaky:
Before this change, I received 401 responses.
After this change, the request was a success.

Example program:

```ruby
require 'date'
require 'json'
require 'workos'

WorkOS.key = ENV.fetch('WORKOS_API_KEY')

event = {
  action_type: 'r',
  action: 'user.login_succeeded',
  actor_id: 'user_01DEQWZNQT8Y47HDPSJKQS1J3F',
  actor_name: 'WorkOS Ruby Gem Test',
  group: 'foo-corp.com',
  latitude: '40.676300',
  longitude: '-73.949200',
  location: '65.215.8.114',
  occurred_at: DateTime.now.iso8601,
  target_id: 'user_01DEQWZNQT8Y47HDPSJKQS1J3F',
  target_name: 'WorkOS Ruby Gem Test'
}

WorkOS::AuditTrail.create_event(event: event)
```

### DIFF
--- a/lib/workos/audit_trail.rb
+++ b/lib/workos/audit_trail.rb
@@ -51,6 +51,7 @@ module WorkOS
       def create_event(event:, idempotency_key: nil)
         request = post_request(
           path: '/events',
+          auth: true,
           idempotency_key: idempotency_key,
           body: event,
         )


blairlunceford (approved):
I recreated the 401 Authentication error, and `auth: true` parameter is needed for the `POST /events` endpoint.

This flag adds the `Authorization` header `Bearer: <WorkOS key>` : https://github.com/workos-inc/workos-ruby/blob/master/lib/workos/client.rb#L50

Is that a correct read cc/ @willmanduffy or @marktran - just in case I'm missing anything.

croaky:
To avoid similar bugs in the future, a larger change requiring more testing would be to reverse the default behavior of `auth: true` to `auth: false` for `WorkOS::Client#post_request` and `WorkOS::Client#get_request`.

By visual reading of the code, the far more common case for callers of `get_request` and `post_request` is `auth: true`. The only exceptions I see are:

* `WorkOS::AuditTrail.create_event` (bug)
* `WorkOS::SSO.profile` (uses alternate auth mechanism in the body instead of headers - intentional?)
* and `spec/lib/workos/base_spec.rb` (test)

Perhaps if those three callsites could be reworked, the `auth` parameter could be removed altogether, and without a major API version bump as internal methods.

willmanduffy (approved):
This looks great @croaky. Thanks for all the detailed feedback on this. We'll definitely look into how we can improve this holistically going forward. I'm going to merge this in now to unblock you today.

croaky:
Sounds good, thanks!

PR #65, opened 2020-12-09
Require JSON

croaky:
While writing an example program for
https://github.com/workos-inc/workos-ruby/pull/64
I noticed an error when not requiring JSON from the standard library:

```
gems/workos-0.9.1/lib/workos/client.rb:65:in `post_request':
undefined method `to_json' for #<Hash:0x00007fa22e11b4f8> (NoMethodError)
```

Example:

```ruby
require 'date'
require 'workos'

WorkOS.key = ENV.fetch('WORKOS_API_KEY')

event = {
  action_type: 'r',
  action: 'user.login_succeeded',
  actor_id: 'user_01DEQWZNQT8Y47HDPSJKQS1J3F',
  actor_name: 'WorkOS Ruby Gem Test',
  group: 'foo-corp.com',
  latitude: '40.676300',
  longitude: '-73.949200',
  location: '65.215.8.114',
  occurred_at: DateTime.now.iso8601,
  target_id: 'user_01DEQWZNQT8Y47HDPSJKQS1J3F',
  target_name: 'WorkOS Ruby Gem Test'
}

WorkOS::AuditTrail.create_event(event: event)
```

I believe the code worked in my Rails app as a side effect of other
dependencies and in the workos-ruby test suite due to `simplecov` gem
depending on `json` gem as a development dependency.

Files with `JSON` or `to_json` need to `require 'json'` for example
programs like the snippets in API reference to work standalone:
https://workos.com/docs/reference/audit-trail/event/publish

Reduced further:

```ruby
{}.to_json # undefined method `to_json' for {}:Hash (NoMethodError)
JSON.parse("{}") # uninitialized constant JSON (NameError)
```

### DIFF
--- a/lib/workos.rb
+++ b/lib/workos.rb
@@ -3,6 +3,7 @@
 
 require 'workos/version'
 require 'sorbet-runtime'
+require 'json'
 
 # Use the WorkOS module to authenticate your
 # requests to the WorkOS API. The gem will read
--- a/lib/workos/audit_trail.rb
+++ b/lib/workos/audit_trail.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 # typed: true
 
+require 'json'
 require 'net/http'
 require 'uri'
 
--- a/lib/workos/audit_trail.rb
+++ b/lib/workos/audit_trail.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 # typed: true
 
-require 'json'
 require 'net/http'
 require 'uri'
 
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 # typed: true
 
-require 'json'
-
 module WorkOS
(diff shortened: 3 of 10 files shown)


blairlunceford (approved):
We already require the JSON gem in the following files:
- https://github.com/workos-inc/workos-ruby/blob/master/lib/workos/connection.rb
- https://github.com/workos-inc/workos-ruby/blob/master/lib/workos/organization.rb
- https://github.com/workos-inc/workos-ruby/blob/master/lib/workos/profile.rb

I don't see an issue in adding it to the others. WDYT @willmanduffy or @marktran ?

willmanduffy (approved):
This definitely works to resolve this issue. @croaky, I'm curious, is your use case also resolved if we put the require inside of `lib/workos.rb` once? That way all files in the `WorkOS` module should get access too.

croaky:
Hi @willmanduffy, definitely! I pushed that change as a follow-up commit to the branch and added some detail about what I believe is mechanically happening to the commit message.

PR #68, opened 2020-12-29
Support client _id and deprecate project_id

blairlunceford:
Clubhouse story: https://app.clubhouse.io/workos/story/4515/ruby-add-client-id-to-sso-package

Ruby SDK updates to support move from projects to environments in WorkOS dashboard.


blairlunceford:
Just got more guidance from Rohan that this shouldn't be a breaking change, and should support both project_id and client_id with a deprecation warning for using client_id. This PR is not ready for review given that info, so I will update the changes with the new guidance.

blairlunceford:
@marktran @willmanduffy I had some questions on the Rubocop errors, and wanted to see if this is the correct change with the move to Dashboard2.

The Rubocop errors are for `Method has too many lines.` and `Avoid parameter lists longer than 5 parameters.`. I can extract the project_id check/deprecation warning into it's own method to address the `Method has too many lines.` error. I'm not sure what to do about the `Avoid parameter lists longer than 5 parameters.` error in the interim while we support both `project_id` and `client_id` - any ideas?

marktran:
> @marktran @willmanduffy I had some questions on the Rubocop errors, and wanted to see if this is the correct change with the move to Dashboard2.
> 
> The Rubocop errors are for `Method has too many lines.` and `Avoid parameter lists longer than 5 parameters.`. I can extract the project_id check/deprecation warning into it's own method to address the `Method has too many lines.` error. I'm not sure what to do about the `Avoid parameter lists longer than 5 parameters.` error in the interim while we support both `project_id` and `client_id` - any ideas?

@blairlunceford Let's just bump the Rubocop config so the limit can handle these cases. The defaults are pretty strict and I think we'll be ok with a bit more tolerance.

blairlunceford:
@marktran thanks! I updated the Rubocop limits for both `MethodLength` and `ParameterLists` 👍

lib/workos/sso.rb
```diff
@@ -51,20 +53,32 @@ class << self
...
       def authorization_url(
-        project_id:, redirect_uri:, domain: nil, provider: nil, state: ''
+        redirect_uri:,
+        project_id: nil,
+        client_id: nil,
+        domain: nil,
+        provider: nil,
+        state: ''
       )
+        unless project_id.nil?
```
marktran, line 72: double negative
```suggestion
        if project_id
```

lib/workos/sso.rb
```diff
@@ -97,10 +113,22 @@ def authorization_url(
...
-      def profile(code:, project_id:)
+      sig do
+        params(
+          code: String,
+          project_id: T.nilable(String),
+          client_id: T.nilable(String),
+        ).returns(WorkOS::Profile)
+      end
+      def profile(code:, project_id: nil, client_id: nil)
+        unless project_id.nil?
```
marktran, line 124:
```suggestion
        if project_id
```

spec/lib/workos/sso_spec.rb
```diff
@@ -113,6 +113,49 @@
...
+          redirect_uri: 'foo.com/auth/callback',
+          state: {
+            next_page: '/dashboard/edit',
+          }.to_s,
+        }
+      end
+      it 'raises a deprecation warning' do
+        expect do
+          described_class.authorization_url(**args)
+        end.to output(
```
marktran, line 131: ❤️

marktran (approved):
looks great

PR #71, opened 2021-01-28
Add GET /connections/:id endpoint

blairlunceford:
Adding a GET /connections/:id endpoint


spec/lib/workos/sso_spec.rb
```diff
@@ -430,4 +430,42 @@
       end
     end
   end
+
+  describe '.get_connection' do
+    before(:all) do
+      WorkOS.key = 'sk_m5CW69Xijbkb7pE8ekNhxXfY9'
```
marktran, line 436: Is this from a real environment?

PR #77, opened 2021-03-12
Add blocks of code setting WorkOS.key per test

willmanduffy:
* The majority of the time our test suite does not need to make live
  requests against the API. We want to avoid this step where we set
  API keys in setup blocks to make requests. If you need to make a
  request (for example, setting up a new VCR cassette) run:
  WORKOS_API_KEY=<key> rspec

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -48,5 +48,6 @@ RSpec.configure do |config|
     end
   end)
 
+  config.before(:all) { WorkOS.key ||= '' }
   config.before(:each) { VCR.turn_on! }
 end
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -48,5 +48,6 @@ def stub_request(*args)
     end
   end)
 
+  config.before(:all) { WorkOS.key ||= '' }
```
willmanduffy, line 51: Usually we don't really need a key.. it just needs to not be `nil`. This is fine enough to run the test suite. If we want an API key this will allow that with

```
WORKOS_API_KEY=<api> rspec
```

blairlunceford (approved):

PR #82, opened 2021-04-23
Support pagination from the WorkOS API

blairlunceford:
Updates the list connection, list directory, list directory groups and list directory users methods to use the ListStruct, which has the listMetadata for supporting pagination.


blairlunceford:
@willmanduffy @marktran let me know if the logic to expose the pagination is correct here. I used the List Organizations method as a guide. Thanks!

lib/workos/group.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+# typed: true
+
+module WorkOS
+  # The Group class provides a lightweight wrapper around
+  # a WorkOS Group resource. This class is not meant to be instantiated
+  # in user space, and is instantiated internally but exposed.
+  class Group
```
willmanduffy, line 8: How are we naming this in other SDKs? I feel this should be named `DirectoryGroup` to a.) be more specific and consistent with our API b.) clear up the `Group` name in case we ever name something in our API that.

willmanduffy (commented):
I think this looks good with the one big comment. We'll want to release this as a breaking change in our API.

lib/workos/group.rb
blairlunceford: That's a good point - I think it's a bit all over the map, but I like the more descriptive naming. I'll update `Group` and `User` to `DirectoryGroup` and `DirectoryUser` respectively, if that sounds good.

willmanduffy: That sounds great.

maxdeviant:
> I think this looks good with the one big comment. We'll want to release this as a breaking change in our API.

Breaking change of the API? Or the SDK?

willmanduffy:
SDK @maxdeviant , apologies for the confusion in my wording!

blairlunceford:
@willmanduffy so it should be released as a major version to indicate the changes?

It does change the format of the data being returned for the list connections, list directories, list directory groups and list directory users endpoints. However, it's a better format that allows for pagination.

willmanduffy (approved):

PR #89, opened 2021-05-04
Update connection status to state

blairlunceford:
- Support both connection status to state

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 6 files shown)


willmanduffy (commented):
This is going to be a breaking change so we should bump to 0.12 I think

blairlunceford:
@willmanduffy we got more guidance from Marshall and we're actually going to support both fields currently, and then deprecate `status` in the future. I'll create a new patch version PR once the three open PRs are merged.

PR #94, opened 2021-05-11
Remove support for deprecated `project_id` parameter

maxdeviant:
This PR removes support for the deprecated `project_id` parameter.

This is a breaking change that will be part of `v1.0.0`.

Resolves SDK-162.


marktran (approved):

maxdeviant:
@marktran Pretty sure this is the first Ruby I've written (or should I say, deleted) in years.

marktran:
> @marktran Pretty sure this is the first Ruby I've written (or should I say, deleted) in years.

🥳

PR #96, opened 2021-05-12
Rename `profile` to `profile_and_token`

maxdeviant:
This PR renames the `profile` method to `profile_and_token`.

This is a breaking change that will be part of `v1.0.0`.

Resolves SDK-155.


marktran (approved):
👍 quality ruby

PR #99, opened 2021-05-19
Add WorkOS::SSO#get_profile

willmanduffy:
* Adds a method to retreive a profile with an access token.
* Adds to ability to use access tokens as a Bearer token in place of
  an API key.

### DIFF
--- a/lib/workos/profile.rb
+++ b/lib/workos/profile.rb
@@ -48,22 +48,20 @@ module WorkOS
 
     private
 
-    # rubocop:disable Metrics/AbcSize
     sig { params(json_string: String).returns(WorkOS::Types::ProfileStruct) }
     def parse_json(json_string)
       hash = JSON.parse(json_string, symbolize_names: true)
 
       WorkOS::Types::ProfileStruct.new(
-        id: hash[:profile][:id],
-        email: hash[:profile][:email],
-        first_name: hash[:profile][:first_name],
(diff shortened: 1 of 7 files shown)


lib/workos/profile.rb
```diff
@@ -48,22 +48,20 @@ def to_json(*)
...
       WorkOS::Types::ProfileStruct.new(
-        id: hash[:profile][:id],
-        email: hash[:profile][:email],
-        first_name: hash[:profile][:first_name],
-        last_name: hash[:profile][:last_name],
-        connection_id: hash[:profile][:connection_id],
-        connection_type: hash[:profile][:connection_type],
-        idp_id: hash[:profile][:idp_id],
-        raw_attributes: hash[:profile][:raw_attributes],
+        id: hash[:id],
```
willmanduffy, line 56: This was previously linked to the data structure that came from `ProfileAndToken`. I've modified that so the generation of the Profile struct assumes we are receiving only the profile hash. I've also modified the `ProfileAndToken` code to send only the Profile.

maxdeviant: 🙏🏻  Thanks so much! I had originally wanted to do this when I first introduced `ProfileAndToken`.

maxdeviant (approved):

maxdeviant:
@willmanduffy It looks like CI is failing cause it can't find older Ruby versions (1.9.3 and 2.0.0).

PR #104, opened 2021-06-01
Remove stray `}`

teoljungberg:
While upgrading from WorkOS 1.0.0 to 1.2.0 we found a regression where
our `Authorization` header accidentally had included an extra `}`. After
searching throughout our application to find the typo, we realized it
had snuck in upstream.

This was introduced in #99


blairlunceford (approved):

willmanduffy:
Thank you so much for catching this @teoljungberg and my apologies for the inconvenience!

PR #106, opened 2021-06-03
Support error and error description in the body of an error

blairlunceford:
The Ruby SDK currently only supports a `message` in the body of an error, however the API sends an `error` and `error_description` in some error bodies.

We should support both in the error handling.

@maxdeviant this is a first pass at improving the error handling. Let me know if you have thoughts on how to implement it better. Additionally, will all errors that have an `error` in the body also have an `error_description`? Or should we account for the case where there is an `error` but no `error_description`?

### DIFF
--- a/lib/workos/errors.rb
+++ b/lib/workos/errors.rb
@@ -13,12 +13,16 @@ module WorkOS
     sig do
       params(
         message: T.nilable(String),
+        error: T.nilable(String),
+        error_description: T.nilable(String),
         http_status: T.nilable(Integer),
         request_id: T.nilable(String),
       ).void
     end
-    def initialize(message: nil, http_status: nil, request_id: nil)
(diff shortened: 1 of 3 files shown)


lib/workos/errors.rb
```diff
@@ -27,7 +31,12 @@ def initialize(message: nil, http_status: nil, request_id: nil)
     def to_s
       status_string = @http_status.nil? ? '' : "Status #{@http_status}, "
       id_string = @request_id.nil? ? '' : " - request ID: #{@request_id}"
-      "#{status_string}#{@message}#{id_string}"
+      if @error && @error_description
```
maxdeviant, line 34: It's possible that we could have an `error` without and `error_description`, so we should probably account for that case.

maxdeviant (approved):

PR #110, opened 2021-07-02
Fix Thread Safety Issues with `Net::HTTP`

marckysharky:
`Net::HTTP` [is not thread safe](https://stackoverflow.com/questions/3063088/is-rubys-nethttp-threadsafe), therefore memoization will potentially lead to issues when the code is run in a multi threaded environment (e.g Puma server with multiple workers). 

This PR amends the Client behaviour to return a new instance of `Net::HTTP` on each request.

### DIFF
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -9,12 +9,9 @@ module WorkOS
 
     sig { returns(Net::HTTP) }
     def client
-      return @client if defined?(@client)
-
-      @client = Net::HTTP.new(WorkOS::API_HOSTNAME, 443)
-      @client.use_ssl = true
-
-      @client
+      Net::HTTP.new(WorkOS::API_HOSTNAME, 443).tap do |http_client|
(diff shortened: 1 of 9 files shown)


mthadley (approved):
Thanks for the fix, and sorry about the wait!

This seems pretty safe to merge.

At worst, there's slightly more overhead in creating a new `Net::HTTP` client, and at best we are solving any threading issues folks have been hitting. It may also vary by implementation, and this might be more of a concern for anyone using JRuby, etc.

Out of curiosity, I took a look at a few other Ruby-based SDK's. Some use other dependencies like Faraday, and so would defer to them for thread-safety. I saw another much-loved SDK that relied on `Net::HTTP` like ours, and it did guard against thread-safety by using `Thread.current` (or thread-local client instances).

The changes themselves look reasonable, and since the various resources `include Client`, it makes sense to use a shared spec to assert that none of them override the thread-safe behavior.

PR #113, opened 2021-07-13
Add webhook signature and timestamp validation methods

blairlunceford:
First SDK pass at adding webhook signature and timestamp validation methods.

I broke up the two methods to give users options, but we could also combine them into one validation method.

I assume that the application will parse the request for the raw POST body and the signature like:

```
sig_header = request.headers['WorkOS-Signature']
payload = request.body.read
```

And then enter the parsed information as parameters to the methods, along with the Webhook Secret.


lib/workos/webhooks.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+
+        if timestamp < Time.now - tolerance
+          raise WorkOS::SignatureVerificationError.new(
+            message: 'Timestamp outside the tolerance zone',
+          )
+        end
+
+        expected_sig = compute_signature(timestamp: timestamp, payload: payload, secret: secret)
+        unless expected_sig.eql?(signature_hash)
```
mthadley, line 98: I think we want to use a constant-time comparison here, and I don't think `String#eql?` is one.

I know Rails provides [`ActiveSupport::SecurityUtils.secure_compare`](https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html#method-c-secure_compare), but maybe there's an equivalent in the Ruby core library or in the deps we already have.

lib/workos/webhooks.rb
```diff
@@ -0,0 +1,148 @@
...
+  # sig_header = request.headers['WorkOS-Signature']
+  # payload = request.body.read
+  #
+  # The secret is the Webhook Secret from your WorkOS Dashboard
+  # The tolerance is for the timestamp validation
+  #
+  module Webhooks
+    class << self
+      extend T::Sig
+      include Base
```
mthadley, line 19: Do we need to include `Base` here? Seems like we don't use any of its behavior, and the specs still pass without it.

lib/workos/webhooks.rb
```diff
@@ -0,0 +1,148 @@
...
+          sig_header: String,
+          secret: String,
+          tolerance: Integer,
+        ).returns(T::Boolean)
+      end
+      # rubocop:disable Metrics/MethodLength
+      def verify_header(
+        payload:,
+        sig_header:,
+        secret:,
```
mthadley, line 74: Looks like other modules default this to `WorkOS.key!`. Should we do that instead of requiring the secret to be passed in?

lib/workos/webhooks.rb
mthadley: Did a little research, and I think we can use [`OpenSSL.secure_compare`](https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL.html#method-c-secure_compare) here instead of `String.eql?`.

Another alternative is to compare instances of `OpenSSL::HMAC`, [as it implements `==` using a constant time comparison](https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/HMAC.html#3D-3D-method). Either one should do the trick, but maybe the first one is easier and more explicit.

spec/lib/workos/webhooks_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+      end
+      it 'raises an error' do
+        expect do
+          described_class.construct_event(**args)
+        end.to raise_error(
+          WorkOS::SignatureVerificationError,
+          'Unable to extract timestamp and signature hash from header',
+        )
+      end
+    end
```
mthadley, line 114: Small nitpick: Since we only use these `let` variables in one place, what do you think about inlining them, along with their `context`?

```rb
    it 'raises an error with an empty header' do
      expect do
        described_class.construct_event(
          payload: @payload,
          sig_header: '',
          secret: @secret,
        )
      end.to raise_error(
        WorkOS::SignatureVerificationError,
        'Unable to extract timestamp and signature hash from header',
      )
    end
```

lib/workos/webhooks.rb
```diff
@@ -0,0 +1,148 @@
+# frozen_string_literal: true
+# typed: true
+
+require 'json'
+require 'openssl'
+
+module WorkOS
+  # The Webhooks module provides convenience methods for working with the WorkOS webhooks.
+  # You'll need to extract the signature header and payload from the webhook request
+  # sig_header = request.headers['WorkOS-Signature']
```
mthadley, line 10: Maybe an extra newline in this comment to separate the code example from the documentation text.

mthadley (changes requested):
Overall, this looks awesome!

Really happy to see that we'll implement this verification process for our users. I think it'll be great to have these security best-practices built-in to our SDK's.

Only thing that I think we need to change is to use a constant time algorithm when comparing the two digests. See the comment I left there for suggestions.

lib/workos/webhooks.rb
blairlunceford: This is actually the Webhook Secret, not the API key, so I can't use `WorkOS.key!` in this case, and it would depend on which webhook endpoint they are using.

lib/workos/webhooks.rb
blairlunceford: I'm using `OpenSSL.secure_compare` now but getting a Sorbet error. Investigating now, but do you have any insight?

Change:

```
unless OpenSSL.secure_compare(expected_sig, signature_hash)
```

Error:

```
lib/workos/webhooks.rb:96: Method secure_compare does not exist on T.class_of(OpenSSL) https://srb.help/7003
    96 |        unless OpenSSL.secure_compare(expected_sig, signature_hash)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Got T.class_of(OpenSSL) originating from:
    lib/workos/webhooks.rb:96:
    96 |        unless OpenSSL.secure_compare(expected_sig, signature_hash)
                       ^^^^^^^
Errors: 1
```

lib/workos/webhooks.rb
mthadley: Ah, that makes sense. Thanks for clarifying!

lib/workos/webhooks.rb
mthadley: So looks like Sorbet's typings for `OpenSSL` don't include `secure_compare`. But actually, that's a good reminder that older versions of Ruby won't have this method in their standard library (looks to be added in 2.4). So maybe we can implement our own to keep that backwards compatibility.

blairlunceford: I used the ActiveSupport method as a guide, but had to insert a type check to pass sorbet tests:

https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/activesupport/lib/active_support/security_utils.rb#L26

mthadley: > but had to insert a type check to pass sorbet tests

Ah, yeah. They type `String#unpack` as `T.nilable(T.any(Integer, Float, String))`, which is like the union of all the the various types that the `unpack` directives could returned. However, since the code uses `C[Int]`, I think it'll always be an array of integers.

I think the implementation from ActiveSupport is probably already correct, given its wide usage, even if Sorbet has trouble validating it. I'm also not sure if it's worth trying to avoid the overhead of the `is_a?` type-check for each byte in the string.

For example, we could do this:

```rb
      def secure_compare(str_a:, str_b:)
        return false unless str_a.bytesize == str_b.bytesize

        l = T.cast(str_a.unpack("C#{str_a.bytesize}"), T::Array[Integer])

        res = 0
        str_b.each_byte do |byte|
          v = T.must(l.shift)

          res |= byte ^ v
        end

        res.zero?
      end
```

Or, if we want to keep the implementation even closer to the original, just opt-out with `T.unsafe`:

```rb
      def secure_compare(str_a:, str_b:)
        return false unless str_a.bytesize == str_b.bytesize

        l = T.unsafe(str_a.unpack "C#{str_a.bytesize}")

        res = 0
        str_b.each_byte { |byte| res |= byte ^ l.shift }

        res.zero?
      end
```

I don't feel strongly either way. I just wanted to explore what other options Sorbet gave us in order to avoid modifying the original algorithm.

mthadley (approved):

lib/workos/webhooks.rb
blairlunceford: Thanks! I went with the second option - I like sticking as close as possible to the original algorithm 👍

mthadley: Cool! Looks good!

PR #135, opened 2021-12-14
Support organization parameter in Get Authorization URL method

blairlunceford:
Adds `organization` parameter to the Get Authorization URL method. Also adds deprecation warning for `domain`.

### DIFF
--- a/lib/workos/sso.rb
+++ b/lib/workos/sso.rb
@@ -21,23 +21,27 @@ module WorkOS
       # Generate an Oauth2 authorization URL where your users will
       # authenticate using the configured SSO Identity Provider.
       #
-      # @param [String] domain The domain for the relevant SSO Connection
-      #  configured on your WorkOS dashboard. One of provider or domain is
-      #  required
-      # @param [String] provider A provider name for an Identity Provider
-      #  configured on your WorkOS dashboard. Only 'Google' is supported.
-      # @param [String] connection The ID for a Connection configured on
-      #  WorkOS.
-      # @param [String] client_id The WorkOS client ID for the environment
-      #  where you've configured your SSO connection.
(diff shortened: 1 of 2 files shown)


lib/workos/sso.rb
```diff
@@ -67,12 +73,19 @@ def authorization_url(
         domain: nil,
         provider: nil,
         connection: nil,
+        organization: nil,
         state: ''
       )
+        if domain
+          warn '[DEPRECATION] `domain` is deprecated.
+          Please use `organization` instead.'
```
willmanduffy, line 81: Just thinking it'd be nice if had a pre-written block here. Some folks are going to hit this when updating their SDKs and not know how to proceed.

willmanduffy (approved):

lib/workos/sso.rb
blairlunceford: Sorry, I'm not sure I follow. What do you mean by pre-written block? Just more information in the warning message?

willmanduffy: Sorry, sometimes my brain subs in the wrong word in the middle of sentences if I'm thinking about the next sentence while writing 😆 

Was thinking it'd be nice if we had a pre-written migration guide of this upgrade that we could link to from these warnings. All the best deprecation warnings I've ever had have a guide of what to do next. I might not know how to get the `organization` to pass in here.

I know we're not quite there yet in this migration, but thinking this is something we could backfill once written.

maxdeviant: I'm going to look into updating the docs today, so I could probably add a page that we can link to with more information.

blairlunceford: Awesome! I'll go ahead and merge this in, but should I hold off on a new version until we have a docs update and update the warning here?

PR #138, opened 2022-01-14
Fix bug in webhook tolerance check

blairlunceford:
@adamzinder found an issue with how the Time was being generated from the extracted timestamp, resulting in a bug in the tolerance check.

This PR is to fix that bug and deal with Time appropriately with the timestamp and tolerance check.

### DIFF
--- a/lib/workos/webhooks.rb
+++ b/lib/workos/webhooks.rb
@@ -65,7 +65,7 @@ module WorkOS
           tolerance: Integer,
         ).returns(T::Boolean)
       end
-      # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


lib/workos/webhooks.rb
```diff
@@ -86,7 +86,9 @@ def verify_header(
           )
         end
 
-        if timestamp < Time.now - tolerance
+        timestamp_to_time = Time.at(timestamp.to_i / 1000)
```
mthadley, line 89: I confirmed that the timestamp we send is in milliseconds, not seconds, and so this fix makes sense. 👍

mthadley (approved):

PR #144, opened 2022-03-27
Feature/sup 261 add mfa to ruby sdk

adamzinder:
MFA added to SDK with testing coverage


lib/workos/mfa.rb
```diff
@@ -0,0 +1,162 @@
...
+
+        if (type == "totp" and totp_issuer == nil) or (type == "totp" and totp_user == nil)
+            raise ArgumentError, "Incomplete arguments. Need to specify both totp_issuer and totp_user when type is totp"
+        end
+
+        if type == "sms" and phone_number == nil
+            raise ArgumentError, "Incomplete arguments. Need to specify phone_number when type is sms"
+            
+        end
+        # this condition is added as type is sms will return a 500 if content type is application/json
```
pliao28, line 70: I think you may be getting this error because the phone number is invalid. An invalid phone number will return a 500 error. And since you are calling this API with vcr/cassette and not a full mock it will give you 500 back.

pliao28: If this is what it is, we can get rid of the whole separate sms type request

pliao28 (commented):
Structure looks good. I think the 500 error comes from the api calling on a non-real phone number for sms type. Some CI syntax errors that will need to be resolved and nitpick: some files don't have newlines at the end of the file

lib/workos/mfa.rb
adamzinder: good find!  that was resolved the issue and I can would remove the if condition as well and tests still pass

lib/workos/challenge_factor.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+# typed: false
+
+module WorkOS
+  # The ChallngeFactor class provides a lightweight wrapper around
+  # a WorkOS DirectoryUser resource. This class is not meant to be instantiated
+  # in DirectoryUser space, and is instantiated internally but exposed.
+  class ChallengeFactor
```
robframpton, line 8: This would be more consistent with what were calling these structures in the other SDKs.

```suggestion
  class Challenge
```

lib/workos/types/challenge_factor_struct.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# typed: strict
+
+module WorkOS
+  module Types
+    # This ChallgengeFactorStruct acts as a typed interface
+    # for the Factor class
+    class ChallengeFactorStruct < T::Struct
```
robframpton, line 8:
```suggestion
    class ChallengeStruct < T::Struct
```

spec/lib/workos/mfa_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+          VCR.use_cassette 'mfa/verify_factor_generic_expired' do
+            expect do
+              described_class.verify_factor(
+                authentication_challenge_id: 'auth_challenge_01FZ4YVRBMXP5ZM0A7BP4AJ12J',
+                code: '897792',
+              )
+            end.to raise_error(WorkOS::InvalidRequestError)
+          end
+        end
+      end
```
robframpton, line 160: So for this case we should add the `code` field to the base error class here: https://github.com/workos/workos-ruby/blob/master/lib/workos/errors.rb#L7

That means we can also remove the `code` and `message` fields from lib/workos/types/verify_factor_struct.rb. Since an exception is getting thrown those fields aren't in use.

lib/workos/challenge_factor.rb
adamzinder: done!

spec/lib/workos/mfa_spec.rb
adamzinder: done!

lib/workos/challenge_factor.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
```
robframpton, line 1: Looks like we have an orphaned file from the rename.

robframpton (approved):

robframpton (approved):

PR #150, opened 2022-04-15
handles instances where valid is false in the reponse with tests

adamzinder:


lib/workos/types/verify_factor_struct.rb
```diff
@@ -7,9 +7,9 @@ module Types
...
     class VerifyFactorStruct < T::Struct
       const :challenge, T.nilable(T::Hash[Symbol, Object])
-      const :valid, T.nilable(TrueClass)
+      const :valid, T.any(TrueClass, FalseClass)
       const :code, T.nilable(String)
       const :message, T.nilable(String)
     end
   end
-end
+end
```
robframpton, line 15: We'll need to add this newline back to remove the Robocop warning that's failing CI.

robframpton (approved):

lib/workos/types/verify_factor_struct.rb
adamzinder: done!

lib/workos/types/verify_factor_struct.rb
```diff
@@ -7,7 +7,7 @@ module Types
     # for the Factor class
     class VerifyFactorStruct < T::Struct
       const :challenge, T.nilable(T::Hash[Symbol, Object])
-      const :valid, T.nilable(TrueClass)
+      const :valid, T.any(TrueClass, FalseClass)
```
marktran, line 10: Does `T::Boolean` work here @adamzinder @robframpton?

spec/lib/workos/mfa_spec.rb
```diff
@@ -135,6 +135,17 @@
         end
       end
     end
+    context 'verify generic otp invalid response' do
```
marktran, line 138: Can we add newlines around all of these `context` blocks, @adamzinder?

Issue #147, opened 2022-04-07
When working with users from a directory, list_users and get_user return different objects.


digerata:
Here is what we see:

__List Users:__

```ruby
irb(main):110:0> WorkOS::DirectorySync.list_users(directory: directory_id, limit: 1).data.first.username
=> "redacted@redacted.com"
irb(main):111:0> WorkOS::DirectorySync.list_users(directory: directory_id, limit: 1).data.first["username"]
Traceback (most recent call last):
        1: from (irb):111
NoMethodError (undefined method `[]' for #<WorkOS::DirectoryUser:0x00005577e10c52e8>)
```

Notice that I could take the first user in the list and then access their username with dot notation. If I tried it as a hash key then it fails.

__Get a User:__

```ruby
irb(main):106:0> WorkOS::DirectorySync.get_user(workos_user_id).username
Traceback (most recent call last):
        2: from (irb):106
        1: from (irb):106:in `rescue in irb_binding'
NoMethodError (undefined method `username' for #<Hash:0x00005577e1b665a8>)
irb(main):107:0> WorkOS::DirectorySync.get_user(workos_user_id)["username"]
=> "redacted@redacted.com"
```

Notice that I retrieve a user and I cannot use dot notation. I have to treat it as a hash.

__Expected:__

I would expect that a user from a list and a user from a get would behave the same way. Either both are accessed through hash values or both are accessed via attributes.

__Why is this an issue?__

We cannot reuse code for "syncing" a user via list and via a lookup.

ams11:
https://github.com/workos/workos-ruby/pull/148 addresses the problem - leaving this issue open for now, until an updated gem with the fix is available.

ams11:
This is now fixed, and a new version of the gem, containing the fix, is available 🎉 - https://rubygems.org/gems/workos/versions/2.2.1.

@digerata, thanks for reporting the issue! 🙇‍♂️ With the current changes, both end points will now always return objects (Array of objects for the list end points), so the preferred method for accessing all the attributes is the dot notation (i.e. `user.name`). For backwards compatibility, we've kept support for `[]` access, so `user["name"]` will also work (for now), but it will be deprecated in a future release, so, I'd try and move away from that style of access.

PR #152, opened 2022-04-15
Add hash support to all models that represent parsed objects

ams11:
A followup to https://github.com/workos/workos-ruby/pull/148 - in https://github.com/workos/workos-ruby/pull/148, we added explicit `to_h` methods to `DirectoryUser` and `DirectoryGroup` models - the two models that were affected by the bug reported in https://github.com/workos/workos-ruby/issues/147. Those methods allow converting the classes into a Hash, which customers may be relying on, since the `get_user` and `get_group` end points were previously returning a Hash.

In order to keep the functionality consistent across the rest of our models, adding a module that defines a `to_h` method (which just calls `to_json`) and including this module in each of our models (I went with the ones that also define a `to_json` method, since I figure those are all the ones that behave similarly to `DirectoryUser` and `DirectoryGroup` - did I miss any?). This way, if any customers were relying on treating the results of the `get_user` and `get_group` calls as a Hash, there'll be able to [fairly] easily update their code (by simply inserting a `to_h`)

💓 to @mthadley for helping formulate the plan for both #147 and this PR.


lib/workos/hash_provider.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+# typed: true
+
+module WorkOS
+  # Module to include an explicit method for converting a model into a Hash containing
+  # its attributes. Default implementation will simply call to_json. Individual classes
+  # may override.
+  module HashProvider
+    include Kernel
```
mthadley, line 9: Do we need to include `Kernel`?

My understanding is that `Kernel` methods are already available everywhere, and so including this module doesn't do anything.

mthadley, line 9: Do we need to include `Kernel`?

My understanding is that `Kernel` methods are already available everywhere, and so including this module doesn't do anything.

ams11: Ahh, I was trying to remember why I'd included that... I can't remember what it was specifically anymore, but it was for a linter error - it wanted an explicit reference to the definition of some method that it couldn't find here in this module. Probably the `raise` call actually... Let me try and run in through the linter again without `Kernel` and see what it complains about.

mthadley: Ah, I see. I found a description of the error in the Sorbet docs: https://sorbet.org/docs/error-reference#7003

I guess they recommend doing this, since there's no guarantee that `raise` will be there if the including object inherits directly from `BasicObject`? [The Ruby docs seem to also recommend this when calling Kernel methods in the context of `BasicObject`](https://ruby-doc.org/core-3.0.2/BasicObject.html).

Seems like a corner case, since most classes will inherit from `Object` which includes `Kernel` by default. But, seems fine to follow the advice from Sorbet here.

TLDR: Nevermind, looks fine to me!

mthadley (approved):

lib/workos/hash_provider.rb
ams11: > Seems like a corner case, since most classes will inherit from Object which includes Kernel by default. But, seems fine to follow the advice from Sorbet here.

Yeah, we can also just disable the Sorbet warning for that line, or call `Kernel.raise` there to be more explicit, but 🤷‍♂️, I don't feel like there's all that much of a difference between any of those options in the end, and [technically, for a very, very edge case of a class that inherits directly from `BasicObject`] including `Kernel` here is safest, so 👍 to just keeping it this way.

> I support another way to implement this is with a base-class, like class Organization < Entity.

Yeah, I considered that also, and honestly don't have a strong preference between the two. Including a module feels just a little bit more flexible, but not by much. I also recall running into some sort of an issue (which I decided not to track down) when I'd tried the base class option, so yeah, as long as you don't feel too strongly about changing this, I'd rather just keeping it as a module to be included.

PR #161, opened 2022-06-10
Add getter method for directory user primary email

blairlunceford:
Adds method on the directory user class to get the primary email out of the email array.

### DIFF
--- a/lib/workos/directory_user.rb
+++ b/lib/workos/directory_user.rb
@@ -47,7 +47,7 @@ module WorkOS
       }
     end
 
-    def get_primary_email(*)
+    def primary_email
       return nil if emails.empty?
 
       emails.each do |email|
--- a/lib/workos/directory_user.rb
+++ b/lib/workos/directory_user.rb
@@ -48,12 +48,8 @@ module WorkOS
     end
 
     def primary_email
-      return nil if emails.empty?
-
-      primary_email = emails.find { |email| email[:primary] }
+      primary_email = (emails || []).find { |email| email[:primary] }
       return primary_email[:value] if primary_email
-
-      nil
     end
 
     private
--- a/lib/workos/directory_user.rb
+++ b/lib/workos/directory_user.rb
@@ -50,9 +50,8 @@ module WorkOS
     def primary_email
       return nil if emails.empty?
 
-      emails.each do |email|
-        return email[:value] if email[:primary]
-      end
+      primary_email = emails.find { |email| email[:primary] }
+      return primary_email[:value] if primary_email
 
       nil
     end
--- a/lib/workos/directory_user.rb
+++ b/lib/workos/directory_user.rb
@@ -47,6 +47,16 @@ module WorkOS
       }
     end
 
+    def get_primary_email(*)
+      return nil if emails.empty?
+
+      emails.each do |email|
+        return email[:value] if email[:primary]
+      end
+
+      nil
+    end
+
     private
 
     sig do
--- a/spec/lib/workos/directory_user_spec.rb
+++ b/spec/lib/workos/directory_user_spec.rb
@@ -7,21 +7,28 @@ describe WorkOS::DirectoryUser do
     context 'with one primary email' do
       it 'returns the primary email' do
         user = WorkOS::DirectoryUser.new('{"object":"directory_user","id":"directory_user_01FAZYNPC8M0HRYTKFP2GNX852","directory_id":"directory_01FAZYMST676QMTFN1DDJZZX87","idp_id":"6092c280a3f1e19ef6d8cef8","username":"logan@workos.com","emails":[{"primary":true,"value":"logan@workos.com"}, {"primary":false,"value":"logan@gmail.com"}],"first_name":"Logan","last_name":"Gingerich","state":"active","raw_attributes":{},"custom_attributes":{},"groups":[]}')
-        expect(user.get_primary_email).to eq('logan@workos.com')
+        expect(user.primary_email).to eq('logan@workos.com')
+      end
+    end
+
+    context 'with multiple primary emails' do
+      it 'returns the first email marked as primary' do
+        user = WorkOS::DirectoryUser.new('{"object":"directory_user","id":"directory_user_01FAZYNPC8M0HRYTKFP2GNX852","directory_id":"directory_01FAZYMST676QMTFN1DDJZZX87","idp_id":"6092c280a3f1e19ef6d8cef8","username":"logan@workos.com","emails":[{"primary":true,"value":"logan@workos.com"}, {"primary":true,"value":"logan@gmail.com"}],"first_name":"Logan","last_name":"Gingerich","state":"active","raw_attributes":{},"custom_attributes":{},"groups":[]}')
+        expect(user.primary_email).to eq('logan@workos.com')
       end
     end
 
     context 'with no primary emails' do
       it 'returns nil' do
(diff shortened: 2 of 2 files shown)


lib/workos/directory_user.rb
```diff
@@ -47,6 +47,16 @@ def to_json(*)
       }
     end
 
+    def get_primary_email(*)
```
ams11, line 50: A couple of minor things here 
- why don't we just call this method `primary_email`, I think the `get` can be implicit.
- do we need the splat `(*)` parameter? I don't think we need any parameters for this method? (🤔 I see we have the `*` ☝️ in `to_json` also... is it actually doing something that I'm not thinking of? Or can we just get rid of the splats everywhere?)
```suggestion
    def primary_email
```

ams11 (commented):
👍 to adding the method here! I had a couple of comments, nothing really major, though I think updating the method name and parameters would be worthwhile...

lib/workos/directory_user.rb
```diff
@@ -47,6 +47,16 @@ def to_json(*)
...
+      return nil if emails.empty?
+
+      emails.each do |email|
+        return email[:value] if email[:primary]
+      end
+
+      nil
```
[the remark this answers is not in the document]
blairlunceford: So we return the value of the primary email, so I'm not sure if `find` will be more concise. But I might be thinking about this the wrong way.

I played around with something like:

```
primary_email = emails.find { |email| email[:primary] }
return primary_email if primary_email
nil
```

But not sure if that's preferred over what I have, or if something else makes more sense.

Thanks!

spec/lib/workos/directory_user_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+# typed: false
+
+describe WorkOS::DirectoryUser do
+  # rubocop:disable Layout/LineLength
+  describe '.get_primary_email' do
+    context 'with one primary email' do
```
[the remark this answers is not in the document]
blairlunceford: 👍

lib/workos/directory_user.rb
blairlunceford: 👍 - yeah I'm not sure of the splat either.

lib/workos/directory_user.rb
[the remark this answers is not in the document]
ams11: Ah, I forgot that we had to actually extract `[:value]` from the `email` object 🤦‍♂️... so, maybe it doesn't really matter then. You could, in theory, use the `&` and do `emails.find { |email| email[:primary] }&.[](:value)`, but that's starting to look pretty messy, so not sure if it's really worth it.

Maybe we just go with the code you have ☝️?
```
primary_email = emails.find { |email| email[:primary] }
return primary_email[:value] if primary_email
nil
```
It still doesn't really matter, but I do [slightly] prefer using `find` over `each`, when we really are trying to find something.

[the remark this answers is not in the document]
ams11: 🤔 `emails.find { |email| email[:primary] }&.[]` might also not work with older version of Ruby/Rails, which I think we still support(?). I can't recall when the `&` came to be, but I think it's relatively recent... so, maybe just another vote for your version of the code ☝️ :).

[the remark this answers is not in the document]
blairlunceford: Sounds good!

lib/workos/directory_user.rb
```diff
@@ -47,6 +47,15 @@ def to_json(*)
       }
     end
 
+    def primary_email
+      return nil if emails.empty?
+
+      primary_email = emails.find { |email| email[:primary] }
+      return primary_email[:value] if primary_email
```
ams11, line 54: ❤️ :shipit:

willmanduffy: We can also maybe shorten this up a bit I think, if we're interested. A couple notes:

1. Ruby will implicity return a nil so need to specify.
2. We don't need to check if emails is empty because an empty array will return nil if nothing is found with `find`

I believe we could write:

```rb
def primary_email(emails)
  (emails || []).find { |email| email[:primary] }
end
```

Some quick tests:

```
irb(main):002:0> emails = [{ email: 'foo@foo-corp.com', primary: true }, { email: 'foo@gmail.com' }]
=> [{:email=>"foo@foo-corp.com", :primary=>true}, {:email=>"foo@gmail.com"}]
irb(main):003:0> empty_emails = []
=> []
irb(main):004:0> nil_emails = nil
=> nil
irb(main):005:0> def primary_email(emails); (emails || []).find { |email| email[:primary] }; end
=> :primary_email
irb(main):006:0> primary_email(emails)
=> {:email=>"foo@foo-corp.com", :primary=>true}
irb(main):007:0> primary_email(empty_emails)
=> nil
irb(main):008:0> primary_email(nil_emails)
=> nil
```

willmanduffy: (We also don't need the `|| []` if we're sure emails is never `nil`)

lib/workos/directory_user.rb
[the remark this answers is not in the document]
willmanduffy: Also just for the curious, the safe navigation operator (`&.`) was Ruby 2.3. About 7 years old now.

lib/workos/directory_user.rb
ams11: I think there's some question as to whether this line is doing quite what we want:

```
irb(main):006:0> primary_email(emails)
=> {:email=>"foo@foo-corp.com", :primary=>true}
```

If we're ok with returning the entire email Hash here, then it works - if we want to return just the email address, then we end up needing to extract it from the hash, which gets us back to being kinda messy (and how we'd landed on the solution ☝️ in the first place - https://github.com/workos/workos-ruby/pull/161#discussion_r894883923). But if we're ok returning the entire Hash, then I'm definitely 👍 for this shorter/cleaner version.

My (entirely personal) preference is to keep `return nil if emails.empty?` at the top of the method - just makes it [visually] more clear what will happen if we have an empty array. I suppose we actually need to do `emails.blank?` though if we do need to worry about `email` being nil (`nil.empty?` is a 💥 😭).

blairlunceford: Thanks! I added those changes 👍 and thanks for the Ruby knowledge! I think we verify here that emails is an array: https://github.com/workos/workos-ruby/blob/main/lib/workos/types/directory_user_struct.rb#L11
but I just kept the check.

lib/workos/directory_user.rb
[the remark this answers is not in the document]
ams11: For extra curiosity, do we publish gems that would be compatible with Ruby < 2.3?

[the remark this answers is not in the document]
blairlunceford: We test older versions than that per the Semaphore file: https://github.com/workos/workos-ruby/blob/main/.semaphore/semaphore.yml#L37

lib/workos/directory_user.rb
blairlunceford: Yeah as a follow up - we would verify that the emails attribute is an array before creating the directory user object, so it shouldn't be `nil`. I can revert back to Alex's suggestion. Whatever works best 👍

ams11: Meh 🤷‍♂️, having the `|| []` in there doesn't do any harm, and makes us more future-proof, in case [for some unknown reason!] we decided to change that logic!

blairlunceford: Let me know if the current version looks good, or if we want to iterate further. Thanks!

PR #162, opened 2022-07-11
Make the `Webooks.verify_header` method public

bfad:
This allows people to semantically verify the body without also parsing the JSON data. This is great for quickly verifying the data in the web process and then stashing it to be parsed and processed in a background process.


blairlunceford:
Hi @bfad thanks for this contribution. We decided to make all methods in the Webhooks class public, including `verify_header`. Closing this PR in favor of https://github.com/workos/workos-ruby/pull/171

PR #164, opened 2022-07-13
Add config initializer for key and timeout

blairlunceford:
Adds a config initializer that would allow users to set timeouts on the HTTP client, and throw a new `TimeoutError` if there is a timeout returned from the API.

There are a few different ways we could implement this, and we could break the config into three separate timeouts: one for open timeout, one for read timeout, and one for write timeout.

Also, the issue specifically mentions the Get Profile and Token method, but that doesn't use `execute_request`, so we might consider changing that as well: https://github.com/workos/workos-ruby/blob/main/lib/workos/sso.rb#L150


lib/workos/configuration.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+# typed: false
+
+module WorkOS
+  # Configuration class sets config initializer
+  class Configuration
+    attr_accessor :timeout
```
mthadley, line 7: Gotta love configuration blocks!

If we are going to start using this pattern, seems like we should consolidate other configuration here as well, right? Thankfully, seems like that would only mean `key` for the moment.

If we don't, developers who configure both `key` and `timeout` would need to use two different patterns:

```rb
# Not great

WorkOS.key = 'key_123abc'
WorkOS.configure do |config|
  config.timeout = 120
end

# Better

WorkOS.configure do |config|
  config.key = 'key_123abc'
  config.timeout = 120
end
```

We'd probably also want to remove `Base` in favor of `Configuration`, and deprecate the key-related singleton class methods on the `WorkOS` module (`key`, `key=`, `key!`, etc.). We need to keep them around for backwards compatibility, but we can implement them in terms of the new pattern:

```rb
module WorkOS
  # ...

  def self.key=(value)
    warn '`WorkOS.key=` is deprecated. Use `WorkOS.configure` instead.'

    config.key = value
  end

  def self.key
    warn '`WorkOS.key` is deprecated. Use `WorkOS.config.key` instead.'

    config.key
  end

  # ...
end
```

@blairlunceford What do you think?

mthadley: We could also leave `WorkOS.key` as a shorthand for developers that don't want to set any other options, but it would still be good to define it in terms of `config`, so that there's a single source of truth for configuration in the Gem.

blairlunceford: Thanks! I moved the key to the configuration, and added the deprecation notices.

lib/workos/configuration.rb
```diff
@@ -0,0 +1,18 @@
...
+# typed: false
+
+module WorkOS
+  # Configuration class sets config initializer
+  class Configuration
+    attr_accessor :timeout, :key
+
+    def initialize
+      @timeout = timeout || 60
+      @key = key
```
mthadley, line 11: I think we can remove the calls to the `key` and `timeout` attr readers, since this is happening inside of `initialize` where it wouldn't be possible for them to have a value other than `nil` yet:

```suggestion
      @timeout = 60
```

I think we can also omit the explicit initialization of `key`, since there isn't a reasonable default for it.

lib/workos/configuration.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    def key!
+      key || raise('WorkOS.config.key not set')
```
mthadley, line 15: Let's omit the explicit parens for `raise`, and also wrap the programmatic identifiers with backticks like we do for other errors/warnings:

```suggestion
      key or raise 'WorkOS.config.key not set'
```

lib/workos/configuration.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# typed: false
```
mthadley, line 2: Any reason not to enable `typed` on this file?

```suggestion
# typed: true
```

lib/workos/configuration.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module WorkOS
+  # Configuration class sets config initializer
+  class Configuration
```
mthadley, line 6: Can we add some test coverage for this new class? Just something basic, to ensure that `timeout`, `key`, and `key!` work as expected, since they are a part of the core behavior of the gem.

blairlunceford: Definitely :) I added some tests - let me know if there is missing coverage 👍

spec/lib/workos/configuration_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+describe WorkOS do
+  describe '.configure' do
+    context 'with key and no timeout' do
+      before do
+        WorkOS.configure do |config|
+          config.key = 'example_api_key'
+        end
+      end
+      it 'sets the key and default timeout configuration' do
```
mthadley, line 12: Tiny nitpick: Can we insert a blank line between test setup and the test themselves?

```suggestion
      end
      
      it 'sets the key and default timeout configuration' do
```

mthadley (approved):
LGTM!

mthadley:
> As a follow-up before merging this - Does one config variable applied to all the timeouts work for this first implementation?

I think it's fine for the moment. 

There's probably rare, but surprising behavior with the way this timeout works. For example, since it is applied separately to three different phases of the request, the whole operation can technically take _longer_ than the timeout, without actually raising the timeout error. However, I'm not sure how many users will run into this.

Perhaps there's a better mechanism to apply a global timeout, and while Ruby has a [`Timeout`](https://ruby-doc.org/stdlib-3.0.1/libdoc/timeout/rdoc/Timeout.html) in the standard library, I've heard [bad](https://adamhooper.medium.com/in-ruby-dont-use-timeout-77d9d4e5a001) [things](https://flushentitypacket.github.io/ruby/2015/02/21/ruby-timeout-how-does-it-even-work.html) [about it](https://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/).

PR #165, opened 2022-07-14
Add organization_id to Directory User and Directory Group

willmanduffy:
* Additionally add directory_id to Directory User


blairlunceford (approved):

blairlunceford:
Only plus would be to add to the tests to check that the org ID is being parsed out, but looks good!

PR #169, opened 2022-08-11
Only set write_timeout if Ruby version is greater or equal to 2.6.0

blairlunceford:
`write_timeout` is only supported for Ruby version 2.6.0+ so we should check that before setting in order to make the gem backwards compatible for older Ruby versions.

### DIFF
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -13,7 +13,7 @@ module WorkOS
         http_client.use_ssl = true
         http_client.open_timeout = WorkOS.config.timeout
         http_client.read_timeout = WorkOS.config.timeout
-        http_client.write_timeout = WorkOS.config.timeout
+        http_client.write_timeout = WorkOS.config.timeout if RUBY_VERSION >= '2.6.0'
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/workos/client.rb
```diff
@@ -13,7 +13,7 @@ def client
         http_client.use_ssl = true
         http_client.open_timeout = WorkOS.config.timeout
         http_client.read_timeout = WorkOS.config.timeout
-        http_client.write_timeout = WorkOS.config.timeout
+        http_client.write_timeout = WorkOS.config.timeout if RUBY_VERSION >= '2.6.0'
```
mthadley, line 16: Looks like we do run CI with Ruby versions older than `2.6`. Did we not have any tests that exercise setting this option?

If not, maybe we should add a few real quick while we are here.

blairlunceford: Yeah I wondered about that. I think it might be because we don't have a test for the Client specifically, but we use the Client throughout the SDK, so I thought the error would get thrown elsewhere.

I'll create a spec just for the client and go from there 👍

blairlunceford: Actually, I found the client spec, it is shared because a lot of the other sections use the client. I added some tests there, but also happy to pull them out into a non-shared spec if that's preferred.

mthadley: Seems fine to me!

mthadley (approved):

PR #171, opened 2022-08-12
Uncover private methods in webhooks class

blairlunceford:
In addition to the previous public method in the Webhooks class, `construct_event`, we also uncover the private methods in this class:

`verify_header`
`get_timestamp_and_signature_hash`
`compute_signature`
`secure_compare`


lib/workos/webhooks.rb
```diff
@@ -55,8 +55,25 @@ def construct_event(
...
+      #     payload: "{"id": "wh_123","data":{"id":"directory_user_01FAEAJCR3ZBZ30D8BD1924TVG","state":"active","emails":[{"type":"work","value":"blair@foo-corp.com","primary":true}],"idp_id":"00u1e8mutl6wlH3lL4x7","object":"directory_user","username":"blair@foo-corp.com","last_name":"Lunceford","first_name":"Blair","directory_id":"directory_01F9M7F68PZP8QXP8G7X5QRHS7","raw_attributes":{"name":{"givenName":"Blair","familyName":"Lunceford","middleName":"Elizabeth","honorificPrefix":"Ms."},"title":"Developer Success Engineer","active":true,"emails":[{"type":"work","value":"blair@foo-corp.com","primary":true}],"groups":[],"locale":"en-US","schemas":["urn:ietf:params:scim:schemas:core:2.0:User","urn:ietf:params:scim:schemas:extension:enterprise:2.0:User"],"userName":"blair@foo-corp.com","addresses":[{"region":"CO","primary":true,"locality":"Steamboat Springs","postalCode":"80487"}],"externalId":"00u1e8mutl6wlH3lL4x7","displayName":"Blair Lunceford","urn:ietf:params:scim:schemas:extension:enterprise:2.0:User":{"manager":{"value":"2","displayName":"Kathleen Chung"},"division":"Engineering","department":"Customer Success"}}},"event":"dsync.user.created"}",
```
ams11, line 68: Are you sure you want to put your name in the examples here? It's not exactly a secret - the GitHub profile is there, but if it's in the code, it may well find a way to spread further 🤷‍♂️

spec/lib/workos/webhooks_spec.rb
```diff
@@ -5,98 +5,70 @@
...
-            division: 'Engineering',
-            department: 'Customer Success',
+        groups: [],
+        locale: 'en-US',
+        schemas: [
+          'urn:ietf:params:scim:schemas:core:2.0:User',
+          'urn:ietf:params:scim:schemas:extension:enterprise:2.0:User'
+        ],
+        userName: 'blair@foo-corp.com',
+        addresses: [{
```
ams11, line 50: Are you sure you are not looking to attract a stalker 😛? I'd definitely put something random into the address fields...

ams11 (approved):
👍 looks good, but I'd take the personal information out of the tests and examples...

Do we have similar in other tests and SDK's? Probably worth saving for a separate PR if so 🤷‍♂️

lib/workos/webhooks.rb
blairlunceford: Yeah I think I was just lazy, and pulled a webhook from my testing 🤦‍♀️ Let me see if I can easily change this example payload.

blairlunceford: Though it is just my name and where I live, which is also on my GitHub and LinkedIn profiles - the email is not real

spec/lib/workos/webhooks_spec.rb
blairlunceford: I think I'm too de-sensitized because I live in a small town where everyone knows everyone 😆

lib/workos/webhooks.rb
ams11: Yeah, I doubt it really makes all that much of a difference, but I generally feel like the less PII, the better ... even if it's not 100% identifiable 🤷‍♂️. I suppose we could also spin as setting a good example of how careful we are when handling customer PII data...

PR #174, opened 2022-08-22
Bundle update sorbet

robframpton:
Currently on `main` I'm receiving the following error when running `bundle install`

```
Your lockfile was created by an old Bundler that left some things out.
Because of the missing DEPENDENCIES, we can only install gems one at a time, instead of installing 16 at a time.
You can fix this by adding the missing gems to your Gemfile, running bundle install, and then removing the gems from your Gemfile.
The missing gems are:
* sorbet-static depended upon by sorbet

...

Unfortunately, an unexpected error occurred, and Bundler cannot continue.
```

The install works after running `bundle update --patch`, however I had to run `bundle exec srb tc -a` to fix issues in `sorbet/rbi/sorbet-typed/lib/rainbow/all/rainbow.rbi`.


mthadley (approved):
Seems fine to me!

I've seen errors like this in the past, though I don't remember what we did to fix them. Maybe we also just bumped the versions of the sorbet-related gems.

PR #175, opened 2022-08-22
Add AuditLogs module

robframpton:
Adds `create_event`, `create_export` and `get_export` methods.

Note: currently includes the changes from https://github.com/workos/workos-ruby/pull/174

### DIFF
--- a/lib/workos/audit_logs.rb
+++ b/lib/workos/audit_logs.rb
@@ -24,7 +24,7 @@ module WorkOS
           organization: String,
           event: Hash,
           idempotency_key: T.nilable(String),
-        ).returns(::T.untyped)
+        ).void
       end
       def create_event(organization:, event:, idempotency_key: nil)
         request = post_request(
@@ -38,8 +38,6 @@ module WorkOS
         )
 
         execute_request(request: request)
-
-        nil
       end
 
       # Create an Export of Audit Log Events.
--- a/lib/workos/audit_logs.rb
+++ b/lib/workos/audit_logs.rb
@@ -14,11 +14,7 @@ module WorkOS
       extend T::Sig
       include Client
 
-      # Create an Audit Log event.
-      #
-      # @param [String] organization An Organization ID
-      # @param [Hash] event An event hash
-      # @param [String] idempotency_key An idempotency key
+      # Create an Audit Log Event.
       sig do
         params(
           organization: String,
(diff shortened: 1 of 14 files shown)


spec/lib/workos/audit_logs_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+          context: {
+            location: '1.1.1.1',
+            user_agent: 'Mozilla',
+          },
+        }
+      end
+
+      context 'with idempotency key' do
+        it 'creates an event' do
+          VCR.use_cassette 'audit_logs/create_event_custom_idempotency_key', match_requests_on: %i[path body headers] do
```
robframpton, line 44: On a side note, I noticed that there are many instances of of `VCR.use_cassette` that don't use `match_requests_on` which means it doesn't matter what the mocked requests have defined in their request body. This can lead to very false positive test results.

lib/workos/audit_logs.rb
```diff
@@ -0,0 +1,110 @@
...
+      # @param [Hash] event An Audit Log Event
+      # @param [String] idempotency_key An idempotency key
+      #
+      # @return [nil]
+      sig do
+        params(
+          organization: String,
+          event: Hash,
+          idempotency_key: T.nilable(String),
+        ).returns(::T.untyped)
```
mthadley, line 27: Should this be `void` instead of returning `T.untyped`?

```suggestion
        ).void
```

lib/workos/audit_logs.rb
```diff
@@ -0,0 +1,110 @@
...
+      end
+
```
mthadley, line 29: Let's remove the newline between the signature and the method implementation.

```suggestion
```

lib/workos/errors.rb
```diff
@@ -18,16 +21,27 @@ class WorkOSError < StandardError
...
     end
-    def initialize(message: nil, error: nil, error_description: nil, http_status: nil, request_id: nil, code: nil)
+    def initialize(
+      message: nil,
+      error: nil,
+      error_description: nil,
+      http_status: nil,
+      request_id: nil,
+      code: nil,
+      errors: nil
```
mthadley, line 34: Does it make sense to have both `error` and `errors`?

spec/lib/workos/audit_logs_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+            range_start: '2022-06-22T15:04:19.704Z',
+            range_end: '2022-08-22T15:04:19.704Z',
+          )
+
+          expect(audit_log_export.object).to eq 'audit_log_export'
+          expect(audit_log_export.id).to eq 'audit_log_export_123'
+          expect(audit_log_export.state).to eq 'pending'
+          expect(audit_log_export.url).to eq nil
+          expect(audit_log_export.created_at).to eq '2022-08-22T15:04:19.704Z'
+          expect(audit_log_export.updated_at).to eq '2022-08-22T15:04:19.704Z'
```
mthadley, line 103: What do you think about combining these assertions into a single `have_attributes`?

```suggestion
          expect(audit_log_export).to have_attributes(
            object: 'audit_log_export',
            id: 'audit_log_export_123',
            state: 'pending',
            url: nil,
            created_at: '2022-08-22T15:04:19.704Z',
            updated_at: '2022-08-22T15:04:19.704Z'
          )
```

lib/workos/errors.rb
robframpton: This is a base class for all our error types. `error` and `error_description` are from [OAuth related errors](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_flow_errors.htm&type=5), while `code`, `message` and `errors` are custom props that we've defined for WorkOS errors.

lib/workos/audit_logs.rb
robframpton: I updated the related assertions to `expect(response).to eq T::Private::Types::Void::VOID` which seems iffy, not sure if there is a more standard way of asserting a void response.

mthadley (approved):

PR #186, opened 2023-02-08
Migrate from `srb rbi` to tapioca

drewfradette:
## Description
Sorbet has moved `srb rbi` to [maintenance mode](https://sorbet.org/blog/2022/07/27/srb-tapioca#what-does-maintenance-mode-mean) and now suggests moving to tapioca which does a better job than `srb rbi`.

Migrating to tapioca also resolves an outstanding issue where `sorbet-typed` paths used characters that are [reserved on NTFS partitions](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions) such as `>`, preventing the gem from being used on Windows-based Ruby environments.

Finally, I removed an unnecessary development dependency for `rake`.

## Documentation

```
[ ] Yes
[X] No
```


sorbet/tapioca/config.yml
```diff
@@ -0,0 +1,13 @@
...
+  # exclude:
+  # - gem_name
+  # doc: true
+  # workers: 5
+dsl:
+  # Add your `dsl` command parameters here:
+  #
+  # exclude:
+  # - SomeGeneratorName
+  # workers: 5
```
tom-workos, line 13: should we put something here besides the autogenerated thing?

tom-workos (approved):

sorbet/tapioca/config.yml
drewfradette: There's no further config necessary so I just left it to indicate "untouched". I can delete the comment lines but then we'll just have
```
gem:
dsl:
```
which looks weirder to me 😄

maxdeviant: Yea, I don't think it hurts to leave the template comments.

PR #202, opened 2023-07-05
Add Devbox

AshGodfrey:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,17 @@
+{
+  "packages": [
+    "ruby@3.0.2",
+    "bundler@latest",
+    "rubocop@0.93.1"
+  ],
+  "shell": {
+    "init_hook": [
+      "sudo gem install rails"
+    ],
+    "scripts": {
+      "test": [
+        "bundle exec rspec"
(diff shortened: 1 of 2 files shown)


devbox.json
```diff
@@ -0,0 +1,16 @@
+{
+  "packages": [
+    "ruby@3.0.2",
+    "rubocop@0.93.1"
+  ],
+  "shell": {
+    "init_hook": [
+      "sudo gem install rails"
```
maxdeviant, line 8: Do we need to install the `rails` gem?

Also, is there a way we can do this without `sudo`?

maxdeviant: I'm thinking we can just use `bundle install` here:

```suggestion
      "bundle install"
```

Which is what we do in CI: https://github.com/workos/workos-ruby/blob/41b7ca022b320469b459b3c68cb7e3567f2c38ec/.semaphore/semaphore.yml#L16-L17

maxdeviant (approved):

drewfradette (approved):

PR #203, opened 2023-07-06
Lazily load configuration env vars

brockspratlen:
## Description
Previously, we were grabbing the value of the `WORKOS_API_HOSTNAME` and `WORKOS_API_KEY` env vars at compile time. For folks using something like dotenv, if dotenv wasn't being loaded before WorkOS then the WorkOS gem wouldn't see the env vars later set by dotenv.

Using lazy loading here helps to ensure that we only capture the environment variables when we actually need them.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/configuration.rb
+++ b/lib/workos/configuration.rb
@@ -4,7 +4,7 @@
 module WorkOS
   # Configuration class sets config initializer
   class Configuration
-    attr_accessor :timeout, :key
+    attr_accessor :api_hostname, :timeout, :key
 
     def initialize
       @timeout = 60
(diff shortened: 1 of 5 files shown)


lib/workos/configuration.rb
```diff
@@ -4,7 +4,7 @@
 module WorkOS
   # Configuration class sets config initializer
   class Configuration
-    attr_accessor :timeout, :key
+    attr_accessor :api_hostname, :timeout, :key
```
mthadley, line 7: Moving this to `Configuration` seems much nicer! 👍

mthadley (approved):
This PR is still in draft, but unless you were planning on any other changes, this already LGTM!

PR #218, opened 2023-08-25
Add delete_user method to UserManagement module

adamzinder:
## Description

* Deletes a user if a valid user ID is passed into the method.

## Documentation

```
[ ] Yes
```


spec/lib/workos/user_management_spec.rb
```diff
@@ -306,4 +306,27 @@
       end
     end
   end
+  describe '.delete_user' do
+    context 'with a valid id' do
+      it 'returns true' do
+        VCR.use_cassette('user_management/delete_user_with_valid_id') do
```
willmanduffy, line 312: I started naming these into folders for cleanliness. 

`user_management/delete_user/valid` for example.

willmanduffy (approved):

spec/lib/workos/user_management_spec.rb
adamzinder: I just changed the file structure to match this design pattern in this PR and the other PRs

PR #221, opened 2023-08-28
Add authenticate_user_password method to UserManagement module

adamzinder:
## Description
Adds authenticate_user_password as well as AuthenticationResponse object to handle nested user object.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/authentication_response.rb
+++ b/lib/workos/authentication_response.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+# typed: true
+
+module WorkOS
+  # The AuthenticationResponse class represents a User as well as an corresponding
+  # response data that can later be appended on.
+  class AuthenticationResponse
+    include HashProvider
+    extend T::Sig
+
+    attr_accessor :user
+
+    sig { params(authenitcation_response_json: String).void }
+    def initialize(authenitcation_response_json)
+      json = JSON.parse(authenitcation_response_json, symbolize_names: true)
+      @user = WorkOS::User.new(json[:user].to_json)
+    end
+
+    def to_json(*)
+      {
+        user: user.to_json,
+      }
+    end
+  end
+end
--- a/lib/workos/authentication_response.rb
+++ b/lib/workos/authentication_response.rb
@@ -10,9 +10,9 @@ module WorkOS
 
     attr_accessor :user
 
-    sig { params(authenitcation_response_json: String).void }
-    def initialize(authenitcation_response_json)
-      json = JSON.parse(authenitcation_response_json, symbolize_names: true)
+    sig { params(authentication_response_json: String).void }
+    def initialize(authentication_response_json)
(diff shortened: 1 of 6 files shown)


lib/workos/authentication_response.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+# typed: true
+
+module WorkOS
+  # The AuthenticationResponse class represents a User as well as an corresponding
+  # response data that can later be appended on.
+  class AuthenticationResponse
```
willmanduffy, line 7: I wonder if we can just return a hash with a User key instead of a full new class? I feel like we're going to be using this return format beyond just Authentication. When the responses get more complex then we can move them to their own classes?

lib/workos/user_management.rb
```diff
@@ -329,6 +329,43 @@ def verify_email(magic_auth_challenge_id:, code:)
 
         WorkOS::User.new(response.body)
       end
+
+
+      # Authenticates user by email and password.
+      #
+      # @param [String] email The email address of the user.
+      # @param [String] password The password to set for the user.
```
willmanduffy, line 337: Is it setting this password for the user? It's authenticating with it.

spec/lib/workos/user_management_spec.rb
```diff
@@ -356,4 +356,37 @@
       end
     end
   end
+  describe '.authenticate_user_password' do
+    context 'with a valid password' do
+      it 'returns user' do
+        VCR.use_cassette('user_management/authenticate_user_password/valid') do
+          authentication_response = WorkOS::UserManagement.authenticate_user_password(
+            email: 'willman@blips.app',
```
willmanduffy, line 364: Let's leave real emails out of the public SDK test suite 😅 I've been manually editing the VCR cassettes as well to remove this.

lib/workos/authentication_response.rb
```diff
@@ -0,0 +1,25 @@
...
+    include HashProvider
+    extend T::Sig
+
+    attr_accessor :user
+
+    sig { params(authenitcation_response_json: String).void }
```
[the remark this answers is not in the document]
adamzinder: done

spec/lib/workos/user_management_spec.rb
adamzinder: sorry about that!  Just removed the actual email from the VCR and test

lib/workos/authentication_response.rb
adamzinder: I'm happy to do this either way.  I know that for the other SDKs (Node, PHP) a new class was used for the response so it would be easier to build off of in the future, but I was thinking about whether a has with a User key would be easier for now.

adamzinder: After discussing with Willman, for now, since parsing is done in the class and not in the API method, we'll go forward with the AuthenticationResponse class.

PR #225, opened 2023-08-31
Add UserResponse class to UserManagement module

adamzinder:
## Description
Adds UserResponse class and removes MagicAuthChallenge and AuthenticationResponse.  Fixes return types for all methods that return UserResponse now, as well as corresponding tests and fixtures.

## Documentation

```
[ ] Yes
```


lib/workos/user_management.rb
```diff
@@ -300,33 +300,32 @@ def send_verification_email(id:)
...
         )
 
-        WorkOS::MagicAuthChallenge.new(response.body)
+        WorkOS::UserResponse.new(response.body)
       end
 
       # Verifies user email using one-time code that was sent to the user.
       #
-      # @param [String] magic_auth_challenge_id The challenge ID returned from the send verification email endpoint.
+      # @param [String] id The unique ID of the User whose email address will be verified.
```
willmanduffy, line 308: This should be `user_id`

willmanduffy (approved):

PR #229, opened 2023-09-28
Don't include development executables

hsbt:
## Description

`workos` gem includes development executables like `console`, `publish` or etc. So, `workos` will install them into bin directory that is part of PATH environmental variables on users machine or production server.

We should remove them because its only for developers of `workos`

## Documentation

```
[ ] Yes
```


mthadley (approved):
Thanks for noticing this, @hsbt!

You are right: the WorkOS gem isn't intended to be shipping any binaries.

PR #231, opened 2023-11-16
Remove Audit Trail modules

LauraBeatris:
## Description

Removes Audit Trail modules since it's a deprecated API. We should only have Audit Logs here from now on.

## Documentation

```
[ ] Yes
```


LauraBeatris:
This package has `yard` which generates documentation, however, we can see that it was last updated back in 2021. It's not working anymore, and those docs aren't being hosted and consumed since then. 

I thought of updating the docs when removing `AuditTrail` here but would be better to raise another PR to delete `yard` all together with the `docs` folder.

amygdalama (approved):

PR #234, opened 2023-11-17
[WIP] Refactor current `UserManagement` methods according to latest API reference

LauraBeatris:
## Description

[WIP]

## Documentation

```
[X] Yes
```

### DIFF
--- a/lib/workos/user_management.rb
+++ b/lib/workos/user_management.rb
@@ -29,7 +29,7 @@ module WorkOS
       def add_user_to_organization(id:, organization_id:)
         response = execute_request(
           request: post_request(
-            path: "/user_management/#{id}/organizations",
+            path: "/user_management/users/#{id}/organizations",
             body: {
               organization_id: organization_id,
             },
@@ -54,7 +54,7 @@ module WorkOS
       def delete_user(id:)
         response = execute_request(
           request: delete_request(
-            path: "/user_management/#{id}",
+            path: "/user_management/users/#{id}",
             auth: true,
           ),
         )
@@ -77,7 +77,7 @@ module WorkOS
       def reset_password(token:, new_password:)
         response = execute_request(
           request: post_request(
-            path: '/user_management/password_reset',
+            path: '/user_management/users/password_reset',
             body: {
               token: token,
               new_password: new_password,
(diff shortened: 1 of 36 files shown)


lib/workos/user_management.rb
```diff
@@ -167,7 +167,7 @@ def create_user(email:, password: nil, first_name: nil, last_name: nil, email_ve
       end
       def update_user(id:, first_name: nil, last_name: nil, email_verified: nil)
         request = put_request(
-          path: "/users/#{id}",
+          path: "/user_management/#{id}",
```
amygdalama, line 170: this (and the other user CRUD endpoints) should be `/user_management/users/`

lib/workos/user_management.rb
```diff
@@ -513,7 +513,7 @@ def authenticate_user_with_totp(code:, client_id:, pending_authentication_token:
       def enroll_auth_factor(user_id:)
         response = execute_request(
           request: post_request(
-            path: "/users/#{user_id}/auth/factors",
+            path: "/user_management/#{user_id}/auth/factors",
```
amygdalama, line 516: also `/user_management/users/`

amygdalama: (and the below auth factors endpoint too)

LauraBeatris: Good point! I was misreading the API controller, thanks for catching!

lib/workos/user_management.rb
```diff
@@ -77,7 +77,7 @@ def delete_user(id:)
       def reset_password(token:, new_password:)
         response = execute_request(
           request: post_request(
-            path: '/users/password_reset',
+            path: '/user_management/users/password_reset',
```
amygdalama, line 80: oh, not everything should have `/user_management/users`. only the REST-style endpoints for User objects... 

the general approach is:
* User Management routes are all "namespaced" under `/user_management`
* within the User Management namespace, we have some REST-style endpoints for User resources. those are `GET /user_management/users/:id` etc
* there are _also_ some endpoints that are part of the User Management product (so using the namespace), but that aren't a REST endpoint on a User resource (like sending a password reset email)

previously, we were treating `/users` as _both_ a product namespace _and_ a resource, and that's what we're moving away from.

check the spreadsheet and/or the API Reference for what these should be 👀

LauraBeatris: Oh cool, I was in automatic mode and rename everywhere 😅 Going to make the fixes here.

amygdalama: yeah no worries, i figured! thank youuu 🙏

LauraBeatris:
@amygdalama Going to break this PR into a smaller one, starting to refactor methods related to `User` only https://github.com/workos/workos-ruby/pull/235

PR #235, opened 2023-11-18
Update User methods according to latest API reference

LauraBeatris:
## Description

- Updates `get_user`, `list_users`, `create_user`, `update_user` and `delete_user`, according to the API docs reference
- Includes missing `object` property to User response struct
- Updates tests

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/types/user_struct.rb
+++ b/lib/workos/types/user_struct.rb
@@ -6,6 +6,7 @@ module WorkOS
     # This UserStruct acts as a typed interface for the User class
     class UserStruct < T::Struct
       const :id, String
+      const :object, String
       const :email, String
       const :first_name, T.nilable(String)
       const :last_name, T.nilable(String)
--- a/lib/workos/types/user_struct.rb
+++ b/lib/workos/types/user_struct.rb
@@ -6,7 +6,6 @@ module WorkOS
     # This UserStruct acts as a typed interface for the User class
     class UserStruct < T::Struct
       const :id, String
-      const :object, String
       const :email, String
       const :first_name, T.nilable(String)
       const :last_name, T.nilable(String)
--- a/lib/workos/user.rb
+++ b/lib/workos/user.rb
@@ -9,7 +9,7 @@ module WorkOS
     include HashProvider
     extend T::Sig
 
-    attr_accessor :id, :object, :email, :first_name, :last_name, :email_verified,
+    attr_accessor :id, :email, :first_name, :last_name, :email_verified,
                   :created_at, :updated_at
 
     sig { params(json: String).void }
@@ -17,7 +17,6 @@ module WorkOS
       raw = parse_json(json)
 
       @id = T.let(raw.id, String)
-      @object = raw.object
       @email = T.let(raw.email, String)
       @first_name = raw.first_name
       @last_name = raw.last_name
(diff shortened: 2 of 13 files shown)


LauraBeatris:
I see that two endpoints are still pending for changes, so going to update this PR once that lands.

lib/workos/types/user_struct.rb
```diff
@@ -6,6 +6,7 @@ module Types
     # This UserStruct acts as a typed interface for the User class
     class UserStruct < T::Struct
       const :id, String
+      const :object, String
```
amygdalama, line 9: i assume the exclusion here was intentional (we use the `object` to determine which type of object to serialize from the json response, but i don't think we include it in the serialized objects, at least in the SDKs i've been working on). do we include the `object` property in other structs here in the Ruby SDK?

amygdalama: otherwise the path changes look good!

LauraBeatris:
@amygdalama We're still waiting on the PUT & POST methods to be finished internally, right? I could merge this one and raise another PR later to update the interface if necessary.

Or do you prefer to hold this?

lib/workos/types/user_struct.rb
amygdalama: oh ok looks like we are inconsistent. here are some w/o the object property:
https://github.com/workos/workos-ruby/blob/main/lib/workos/types/directory_struct.rb
https://github.com/workos/workos-ruby/blob/main/lib/workos/types/organization_struct.rb
https://github.com/workos/workos-ruby/blob/main/lib/workos/types/event_struct.rb

LauraBeatris: Those are missing, I think we should fix it later!

LauraBeatris: But I think ideally all of the SDKs should return the object property as we also display it on our docs response object.

amygdalama: oh i was using those examples to show that i don't think including `object` is required...

amygdalama:
we don't need to wait on the POST and PUT changes. those can be in a separate PR

lib/workos/types/user_struct.rb
LauraBeatris: Yeah, I'm actually not sure. I see that we display those on our docs, so I don't think there could be a case where developers expect that we return those as well within the response instead of omitting them.  

CleanShot 2023-11-20 at 14 39 05

LauraBeatris: But as you mentioned, we are not that consistent about it. Some SDKs do that, but not in all modules.

amygdalama: yeah, here's how i see it:

* for json responses, like if a developer is integrating without using one of our SDKs, they need the `object` property of the json response to know the type of object
* however, when using an SDK, the developer is interacting with serialized objects, like a User or an Event. they can identify which object they are using by the class/instance type. and in that scenario, the object property is redundant

i think we should leave it out for now

amygdalama: (1) is why `object` needs to be documented in the API Reference (for json/curl code snippets)

LauraBeatris: Your point makes sense to me. Let's stop adding :object to the response structs here then - prob also be mindful of that for the other SDKs.

amygdalama: sounds good, thank you!

amygdalama (approved):

PR #236, opened 2023-11-19
Add `get_authorization_url` to User Management module 

LauraBeatris:
## Description

- Adds `AuthKit` as a valid enum value for the `Provider` type 
- Introduces `get_authorization_url` method on `UserManagement` class
- Introduces tests 

## Documentation

```
[X] Yes
```

### DIFF
--- a/lib/workos/types/provider_enum.rb
+++ b/lib/workos/types/provider_enum.rb
@@ -9,7 +9,7 @@ module WorkOS
       enums do
         Google = new('GoogleOAuth')
         Microsoft = new('MicrosoftOAuth')
-        AuthKit = new ('authkit')
+        AuthKit = new('authkit')
       end
     end
   end
--- a/lib/workos/types/provider_enum.rb
+++ b/lib/workos/types/provider_enum.rb
@@ -9,6 +9,7 @@ module WorkOS
       enums do
         Google = new('GoogleOAuth')
         Microsoft = new('MicrosoftOAuth')
+        AuthKit = new ('authkit')
       end
     end
   end
--- a/lib/workos/types/provider_enum.rb
+++ b/lib/workos/types/provider_enum.rb
@@ -9,7 +9,6 @@ module WorkOS
       enums do
         Google = new('GoogleOAuth')
         Microsoft = new('MicrosoftOAuth')
-        AuthKit = new('authkit')
       end
     end
   end
--- a/lib/workos/user_management.rb
+++ b/lib/workos/user_management.rb
@@ -76,7 +76,7 @@ module WorkOS
           state: T.nilable(String),
         ).returns(String)
       end
-      def get_authorization_url(
+      def authorization_url(
         redirect_uri:,
         client_id: nil,
         domain_hint: nil,
@@ -87,7 +87,7 @@ module WorkOS
         state: ''
       )
 
-        validate_get_authorization_url_arguments(
+        validate_authorization_url_arguments(
           provider: provider,
           connection_id: connection_id,
           organization_id: organization_id,
@@ -659,7 +659,7 @@ module WorkOS
         ).void
       end
 
-      def validate_get_authorization_url_arguments(
+      def validate_authorization_url_arguments(
         provider:,
         connection_id:,
         organization_id:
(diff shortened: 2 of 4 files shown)


spec/lib/workos/sso_spec.rb
```diff
@@ -108,40 +108,6 @@
       end
     end
 
-    context 'with a domain' do
```
LauraBeatris, line 111: This block was duplicated, reason why I'm removing it here.

lib/workos/user_management.rb
```diff
@@ -14,6 +14,89 @@ class << self
...
+          client_id: T.nilable(String),
+          domain_hint: T.nilable(String),
+          login_hint: T.nilable(String),
+          provider: T.nilable(String),
+          connection_id: T.nilable(String),
+          organization_id: T.nilable(String),
+          state: T.nilable(String),
+        ).returns(String)
+      end
+      def authorization_url(
```
amygdalama, line 67: i see this was copied from the existing `authorization_url` for sso, but all the other SDKs include "get" in the name, like `get_authorization_url`. do you think we should change the Ruby SDK to be consistent with those?

LauraBeatris: I'd opt to be consistent with the other SDKs, I like to append the method on top of it.

LauraBeatris: Updated on https://github.com/workos/workos-ruby/pull/236/commits/8e36fe5d94eaa8bd9933600e23b2485b2b951418

LauraBeatris: @amygdalama Will also update the docs!

mthadley: Generally in Ruby people avoiding naming methods like `get_foo` (and `set_foo`), in favor of `foo` and `foo=` (i.e. `obj.foo = 5`.

So i can see why we called it `authorization_url`. Might be more consistent for _us_, but I don't think consistency across SDK's really benefits the consumer (since they'll likely only use one of them).

LauraBeatris: That's right! Attributes accessing methods, correct?

Yeah, I see your point. I don't have a strong opinion about it, I'd be fine to keep it as `authorization_url`.

amygdalama: my b! sounds good to me

lib/workos/types/provider_enum.rb
```diff
@@ -9,6 +9,7 @@ class Provider < T::Enum
       enums do
         Google = new('GoogleOAuth')
         Microsoft = new('MicrosoftOAuth')
+        AuthKit = new('authkit')
```
mthadley, line 12: Technically `authkit` is only valid for `/user_management/authorize`, but this `Provider` is type is shared with `/sso/authorize`.

So I think we probably shouldn't reuse this type for the new method.

LauraBeatris: That's right, let's create a new one - Maybe `UserManagementProvider`?

mthadley: Sounds good! I might even suggest we put it in the new `WorkOS::UserManagement` module, so `WorkOS::UserManagement::Types::Provider`.

LauraBeatris: @mthadley I've updated it on https://github.com/workos/workos-ruby/pull/236/commits/f6d7dc289519be268f3e1a223f5c13c21f46073d

Let me know if it looks good!

lib/workos/user_management.rb
LauraBeatris: Method name updated on https://github.com/workos/workos-ruby/pull/236/commits/6356023878e0c407f86fb9b0f5c29c03a9965360

lib/workos/user_management.rb
```diff
@@ -10,10 +10,105 @@ module WorkOS
...
+          AuthKit = new('authkit')
+        end
+      end
+    end
+
     class << self
       extend T::Sig
       include Client
 
+      PROVIDERS = WorkOS::UserManagement::Types::Provider.values.map(&:serialize).freeze
```
mthadley, line 29: Now that `Types` is defined in the same namespace, we don't need the full module path:

```suggestion
      PROVIDERS = Types::Provider.values.map(&:serialize).freeze
```

mthadley (approved):

PR #240, opened 2023-11-20
Refactor Multi-Factor according to the latest API reference

LauraBeatris:
## Description

Refactoring `enroll_auth_factor` & `list_auth_factors` + updating tests

## Documentation

```
[ ] Yes
```


LauraBeatris:
FYI our API is returning `authentication_factor` & `authentication_challenge` as keys brought the API reference is displaying the response with `factor` & `challenge` properties

CleanShot 2023-11-20 at 17 15 22

lib/workos/user_management.rb
```diff
@@ -503,19 +503,28 @@ def authenticate_user_with_totp(code:, client_id:, pending_authentication_token:
...
+      # @param [String] type The type of the factor to enroll. Only option available is totp.
+      # @param [String] totp_issuer For totp factors. Typically your application
+      #  or company name, this helps users distinguish between factors in authenticator apps.
+      # @param [String] totp_user For totp factors. Used as the account name in authenticator apps.
       #
       # @return WorkOS::AuthenticationFactorAndChallenge
       sig do
         params(
           user_id: String,
+          type: String,
```
amygdalama, line 515: one idea: perhaps introduce an Enum (something like `FactorType` or `AuthFactorType`) which currently has a single value (`'totp'`)?

amygdalama (approved):

lib/workos/user_management.rb
LauraBeatris: I've introduced an enum, inside of the User Management module scope, instead of following that pattern of creating as in the global Types namespace. 

Pretty much the same thing that we did for providers, and we're now validating the auth factor: https://github.com/workos/workos-ruby/pull/240/commits/bad936333c8c048f784a1ea652323a628a5cce24

Then let's keep that in mind if one day we add new auth factors, that enum should be updated as well.

PR #253, opened 2023-12-19
Add password, password_hash and password_hash_type to update_user 

jbvolta:
## Description
Add password, password_hash and password_hash_type to UserManagement.update_user function

https://workos.com/docs/reference/user-management/user/update

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/user_management.rb
+++ b/lib/workos/user_management.rb
@@ -220,7 +220,7 @@ module WorkOS
       # @option update_hash [Boolean] email_verified Whether the user's email address was previously verified.
       # @option update_hash [String] password The user's password.
       # @option update_hash [String] password_hash The user's hashed password.
-      # @option update_hash [String] password_hash_type The algorithm originally used to hash the password. 
+      # @option update_hash [String] password_hash_type The algorithm originally used to hash the password.
       #  Valid values are bcrypt.
       sig do
         params(
--- a/lib/workos/user_management.rb
+++ b/lib/workos/user_management.rb
@@ -214,34 +214,36 @@ module WorkOS
       # Update a user
       #
       # @param [String] id of the user.
-      # @param [String] first_name The user's first name.
-      # @param [String] last_name The user's last name.
-      # @param [Boolean] email_verified Whether the user's email address was previously verified.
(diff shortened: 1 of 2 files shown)


lib/workos/user_management.rb
```diff
@@ -214,24 +214,37 @@ def create_user(email:, password: nil, first_name: nil, last_name: nil, email_ve
...
+            last_name: T.nilable(String),
+            email_verified: T.nilable(T::Boolean),
+            password: T.nilable(String),
+            password_hash: T.nilable(String),
+            password_hash_type: T.nilable(String),
+          },
         ).returns(WorkOS::User)
       end
-      def update_user(id:, first_name: nil, last_name: nil, email_verified: nil)
+      def update_user(id:, update_hash:)
```
mthadley, line 238: Is there a reason we are switching to collecting all of the previous keyword arguments into a new `update_hash` parameter? 

This would be a breaking change. It also seems less idiomatic in Ruby than adding new keyword arguments for the password-related fields.

jbvolta: It's one of our Rubocop rule. We can't have 6 parameters in a function.

Feels ok to ship this breaking change since the current one has not yet been released in a stable version.

mthadley: Ah, gotcha. We have other places where we have 6 or more keyword arguments, like here for SSO:

https://github.com/workos/workos-ruby/blob/f88b01a9ded91723d947038b079e965ff6a73a88/lib/workos/sso.rb#L71-L81

In those cases we explicitly disable the rule with `rubocop:disable Metrics/ParameterLists`, but I think we should instead update our `.rubocop.yml` to exclude keyword arguments entirely from this rule, and then remove those magic comments:

```yml
# .rubocop.yml

Metrics/ParameterLists:
  Max: 6
  CountKeywordArgs: false
```

Then we can go back to adding more keyword arguments to the `update_user` method. Does that seem reasonable?

jbvolta: reasonable! I update the PR.

I like this rule tho...

mthadley (approved):

PR #260, opened 2024-02-06
Add `profile_picture_url` attribute to the `User` object

mthadley:
## Description

Adds the new `profile_picture_url` field to the Ruby SDK `User` object.

## Documentation

```
[x] Yes
```


spec/support/fixtures/vcr_cassettes/user_management/get_user.yml
mthadley, line 1: For now I only updated this cassette so that we at least have some kind of smoke test for when the new field is returned.

I think it would be a larger endeavor to update the rest of the cassettes that have User objects.

LauraBeatris (approved):

PR #268, opened 2024-03-14
Add `impersonator` attr to `AuthenticationResponse`

mthadley:
## Description

Adds Impersonation-related fields to the authentication response class.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/spec/lib/workos/user_management_spec.rb
+++ b/spec/lib/workos/user_management_spec.rb
@@ -420,6 +420,22 @@ describe WorkOS::UserManagement do
           expect(authentication_response.user.id).to eq('user_01H93ZY4F80YZRRS6N59Z2HFVS')
         end
       end
+
+      context 'when the user is being impersonated' do
+        it 'contains the impersonator metadata' do
+          VCR.use_cassette('user_management/authenticate_with_code/valid_with_impersonator') do
+            authentication_response = WorkOS::UserManagement.authenticate_with_code(
+              code: '01HRX85ATQB2MN40K4FZ9C2HFR',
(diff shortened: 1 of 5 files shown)


spec/lib/workos/user_management_spec.rb
```diff
@@ -420,6 +420,22 @@
           expect(authentication_response.user.id).to eq('user_01H93ZY4F80YZRRS6N59Z2HFVS')
         end
       end
+
+      context 'when the user is being impersonated' do
+        it 'contains the impersonator metadata' do
+          VCR.use_cassette('user_management/authenticate_with_code/valid_with_impersonator') do
+            authentication_response = WorkOS::UserManagement.authenticate_with_code(
+              code: '01HRX85ATQB2MN40K4FZ9C2HFR',
+              client_id: 'client_01GS91XFB2YPR1C0NR5SH758Q0',
```
cmatheson, line 429: i think i've seen the client_id get turned into client_123 on other tests (but i don't feel strongly about putting [your?] real one in)

i do wish we had an workos account just for sdk tests and we could just use real client ids all the time in here.

cmatheson (approved):

spec/lib/workos/user_management_spec.rb
mthadley: Yeah, I agree it looks funny at first glance, but `client_id` isn't a secret in OAuth (and appears in the authorization URL for all to see).

So I didn't bother to obfuscate it, but wouldn't mind if we added some kind of normalization of it for these tests.

PR #269, opened 2024-03-18
Add session-related endpoints

cmatheson:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/refresh_authentication_response.rb
+++ b/lib/workos/refresh_authentication_response.rb
@@ -2,8 +2,8 @@
 # typed: true
 
 module WorkOS
-  # The AuthenticationResponse class represents an Authentication Response as well as an corresponding
-  # response data that can later be appended on.
+  # The RefreshAuthenticationResponse contains response data from a successful
+  # `UserManagement.authenticate_with_refresh_token` call
   class RefreshAuthenticationResponse
     include HashProvider
     extend T::Sig
--- a/lib/workos/refresh_authentication_response.rb
+++ b/lib/workos/refresh_authentication_response.rb
@@ -4,31 +4,21 @@
 module WorkOS
   # The AuthenticationResponse class represents an Authentication Response as well as an corresponding
   # response data that can later be appended on.
-  class AuthenticationResponse
+  class RefreshAuthenticationResponse
     include HashProvider
     extend T::Sig
 
-    attr_accessor :user, :organization_id, :impersonator, :access_token, :refresh_token
+    attr_accessor :access_token, :refresh_token
 
     sig { params(authentication_response_json: String).void }
     def initialize(authentication_response_json)
       json = JSON.parse(authentication_response_json, symbolize_names: true)
-      @user = WorkOS::User.new(json[:user].to_json)
-      @organization_id = T.let(json[:organization_id], T.nilable(String))
-      @impersonator =
-        if (impersonator_json = json[:impersonator])
-          Impersonator.new(email: impersonator_json[:email],
-                           reason: impersonator_json[:reason],)
-        end
       @access_token = T.let(json[:access_token], String)
       @refresh_token = T.let(json[:refresh_token], String)
     end
 
     def to_json(*)
       {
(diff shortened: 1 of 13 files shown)


spec/support/fixtures/vcr_cassettes/user_management/authenticate_with_code/valid.yml
```diff
@@ -76,7 +76,7 @@ http_interactions:
           - cloudflare
       body:
         encoding: ASCII-8BIT
-        string: '{"user":{"object":"user","id":"user_01H93ZY4F80YZRRS6N59Z2HFVS","email":"test@workos.app","email_verified":false,"first_name":"Lucille","last_name":"Bluth","created_at":"2023-08-30T19:50:13.214Z","updated_at":"2023-08-30T19:50:13.214Z","user_type":"managed","sso_profile_id":"prof_01H93ZTVWYPAT4RKDSPFPPXH0J"}}'
+        string: '{"user":{"object":"user","id":"user_01H93ZY4F80YZRRS6N59Z2HFVS","email":"test@workos.app","email_verified":false,"first_name":"Lucille","last_name":"Bluth","created_at":"2023-08-30T19:50:13.214Z","updated_at":"2023-08-30T19:50:13.214Z","user_type":"managed","sso_profile_id":"prof_01H93ZTVWYPAT4RKDSPFPPXH0J"},"access_token":"access_token","refresh_token":"refresh_token"}'
```
cmatheson, line 79: i kind of hate vcr and i manually edited/added all of these responses. this is not technically the right way to do this.

lib/workos/refresh_authentication_response.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+# typed: true
+
+module WorkOS
+  # The AuthenticationResponse class represents an Authentication Response as well as an corresponding
+  # response data that can later be appended on.
```
mthadley, line 6: What does "an corresponding response data that can later be appended on" refer to?

lib/workos/user_management.rb
```diff
@@ -554,6 +594,66 @@ def authenticate_with_email_verification(
...
+      end
+
+      # Revokes a session
+      #
+      # @param [String] session_id The session ID can be found in the `sid`
+      #   claim of the access token
+      sig do
+        params(
+          session_id: String,
+        ).returns(NilClass)
```
mthadley, line 625: Should this use [`.void`](https://sorbet.org/docs/sigs#returns--void-annotating-return-types) instead?

lib/workos/refresh_authentication_response.rb
cmatheson: ohhhh... this is a complete copy/paste from AuthenticationResponse. i'll come up with better wording :grimacing:

lib/workos/user_management.rb
cmatheson: ahhh, i knew there must be something like this. will fix.

spec/spec_helper.rb
```diff
@@ -26,6 +26,12 @@
 VCR.configure do |config|
   config.cassette_library_dir = 'spec/support/fixtures/vcr_cassettes'
   config.filter_sensitive_data('<API_KEY>') { WorkOS.config.key }
+  config.filter_sensitive_data('<ACCESS_TOKEN>', :token) do |interaction|
+    JSON.parse(interaction.response.body)['access_token']
+  end
+  config.filter_sensitive_data('<REFRESH_TOKEN>', :token) do |interaction|
+    JSON.parse(interaction.response.body)['refresh_token']
+  end
```
mthadley, line 34: It'd be nice if this was the default and test authors didn't need to remember to opt-in, though I don't know the best way to implement that behavior without it also being inconsistent.

cmatheson: ugh, i agree. i tried just having the handler work without tags but it errors out if you don't return something. (i guess i could return some bogus value?). this seemed to be the "vcr" way to handle this situation.

PR #282, opened 2024-04-11
Make events parameter required for list_events

mattgd:
## Description
Make events parameter required for list_events.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/events.rb
+++ b/lib/workos/events.rb
@@ -21,7 +21,7 @@ module WorkOS
       #  after a provided Event ID.
       #
       # @return [Hash]
-      def list_events(options = {})
+      def list_events(events:, options = {})
         response = execute_request(
           request: get_request(
             path: '/events',
--- a/lib/workos/events.rb
+++ b/lib/workos/events.rb
@@ -21,12 +21,15 @@ module WorkOS
       #  after a provided Event ID.
       #
       # @return [Hash]
-      def list_events(events:, options = {})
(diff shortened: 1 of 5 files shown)


mattgd:
I'm not much of a rubyist. Do I need to update the method signature or comment at all on this options hash to reflect this parameter requirement?
https://github.com/workos/workos-ruby/blob/2a90a0bee142c31b11cfe3d85017f5d505f3f0ec/lib/workos/events.rb#L16-L24

jasonroelofs:
This change doesn't look to be doing what you want. The `InvalidRequestError` would come from making the request with no parameters, not the SDK telling the user they need to add a parameter. If `event` is a required parameter I would probably change the signature of `list_event` to thus:

```ruby
def list_event(event:, options = {})
end
```

This way the user will get a `ArgumentError` when they forget `event` and this should also be backwards compatible to current callers to this function.

jasonroelofs:
Ah, right, that syntax isn't 2.7 compatible. You might need to add extra logic looking for `options[:event]` and throw an ArgumentError if it doesn't exist.

mattgd:
> Ah, right, that syntax isn't 2.7 compatible. You might need to add extra logic looking for `options[:event]` and throw an ArgumentError if it doesn't exist.

Ah okay. Let me move make some changes.

PaulAsjes (approved):

PR #291, opened 2024-04-30
Add `domain_data` to Organizations Create and Update methods to replace `domains`

mthadley:
## Description

Adds a new `domain_data` parameter to be passed to Organization create and update calls, instead of the `domains` parameter which will be soon deprecated.

While I was in here I also cleaned up some noise and warnings when running the specs, and also unified the way we were calling `warn` into a shared `WorkOS::Deprecation` module.

## Documentation

```
[ ] Yes
```


lib/workos/organizations.rb
```diff
@@ -97,12 +110,17 @@ def create_organization(domains:, name:, allow_profiles_outside_organization: ni
       # Update an organization
       #
       # @param [String] organization Organization unique identifier
-      # @param [Array<String>] domains List of domains that belong to the
-      #  organization
+      # @param [Array<Hash>] domain_data List of domain hashes describing an orgnaization domain.
+      # @option domain_data [String] domain The domain that belongs to the organization.
+      # @option domain_data [String] state The state of the domain. "verified" or "pending"
+      # @param [Array<String>] domains List of domains that belong to the organization.
+      #   Deprecated: Use domain_data instead.
```
cmatheson, line 117: is this deprecation notice supposed to be attached to the old `domains` param? (i think we lost that one?)

cmatheson (approved):

lib/workos/organizations.rb
mthadley: Yeah, it should be associated with `domains`:

```ruby
      # @param [Array<String>] domains List of domains that belong to the organization.
      #   Deprecated: Use domain_data instead.
```

Maybe it looks a bit weird since it is on the next line, indented, as a separate "paragraph". Is that the right way to document it in Ruby?

cmatheson: ohhh, i don't know how i wasn't seeing that earlier. yeah this looks right to me :+1:

PR #293, opened 2024-05-03
Add API changes for invitations and Magic Auth

blairlunceford:
## Description

Adds API changes to support sending your own emails for invitations and Magic Auth.

- Adds `accept_invitation_url` to invitation object returned by API
- Adds new endpoints for the Magic Auth API: `get_magic_auth` and `create_magic_auth`
- Deprecates current `send_magic_auth_code` method in favor of `create_magic_auth`

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/26691


spec/lib/workos/user_management_spec.rb
```diff
@@ -618,6 +618,46 @@
...
+        VCR.use_cassette 'user_management/create_magic_auth/valid' do
+          magic_auth = described_class.create_magic_auth(
+            email: 'test@workos.com',
+          )
+
+          expect(magic_auth.id).to eq('magic_auth_01HWXVEWWSMR5HS8M6FBGMBJJ9')
+          expect(magic_auth.email).to eq('test@workos.com')
+        end
+      end
+    end
```
tribble, line 658: Do we generally test other variations? There is also a version of this method that accepts an invitation.

tribble (approved):
Approved!

spec/lib/workos/user_management_spec.rb
blairlunceford: Sometimes - testing is a bit spotty in the SDKs. The invitation_token is optional for this method, and it's only needed in a specific case: sign ups are turned off and there isn't already a user with the email.

I'll add a follow up item to add a test for that case. I need to get my dashboard in a specific state to do that test.

PR #295, opened 2024-05-15
Add organization membership deactivate and reactivate API methods

mattgd:
## Description
Add organization membership deactivate and reactivate API methods.

## Documentation

```
[x] Yes
```

### DIFF
--- a/lib/workos/user_management.rb
+++ b/lib/workos/user_management.rb
@@ -748,7 +748,8 @@ module WorkOS
       #
       # @param [Hash] options
       # @option options [String] user_id The ID of the User.
-      # @option options [String] organization_id Filter Users by the organization they are members of.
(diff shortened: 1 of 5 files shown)


lib/workos/user_management.rb
```diff
@@ -748,7 +748,8 @@ def get_organization_membership(id:)
       #
       # @param [Hash] options
       # @option options [String] user_id The ID of the User.
-      # @option options [String] organization_id Filter Users by the organization they are members of.
+      # @option options [String] organization_id Filter memberships by the organization they are members of.
+      # @option options [Array<String>] statuses Filter memberships by status.
```
mattgd, line 752: Is `Array<String>` correct here?

tribble: Given the API accepts an array, I think that's correct

PR #303, opened 2024-06-05
Add new error types and add data to error response

PaulAsjes:
## Description

- Adds new error types to be more specific
- Adds data to 400 error responses so users can find out exactly what went wrong

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/client.rb
+++ b/lib/workos/client.rb
@@ -86,7 +86,7 @@ module WorkOS
       ].join('; ')
     end
 
-    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Metrics/CyclomaticComplexity:
     def handle_error_response(response:)
       http_status = response.code.to_i
       json = JSON.parse(response.body)
@@ -142,7 +142,7 @@ module WorkOS
         )
       end
     end
-    # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
+    # rubocop:enable Metrics/MethodLength, Metrics/AbcSize, Metrics/CyclomaticComplexity:
 
     private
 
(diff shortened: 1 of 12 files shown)


lib/workos/client.rb
```diff
@@ -86,7 +86,7 @@ def user_agent
       ].join('; ')
     end
 
-    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Metrics/CyclomaticComplexity:
```
cmatheson, line 89: can we get rid of these cops? they are so annoying.

lib/workos/client.rb
```diff
@@ -107,7 +110,7 @@ def handle_error_response(response:)
           request_id: response['x-request-id'],
         )
       when 404
-        raise APIError.new(
+        raise NotFoundError.new(
```
cmatheson, line 113: ahh, these more specific error classes are nice. i guess this will be a breaking change though? (needs a major version bump?).

cmatheson (approved):
looks good! maybe a test that shows that the error info is present on the object is in order?

lib/workos/client.rb
PaulAsjes: Yep afraid so, but it's okay as we have some other changes that are ready to go and require a major version.

lib/workos/client.rb
PaulAsjes: Yes, but I don't think this is the PR for it.

mattgd (approved):

lib/workos/client.rb
cmatheson: oh yeah sorry, that was mostly me griping (should have specified that i didn't intend for you to do that in this PR).

PR #307, opened 2024-06-12
Mark `allow_profiles_outside_organization` as deprecated

mthadley:
## Description

Marks `allow_profiles_outside_organization` as now deprecated. Also includes a few other misc. fixes:

* We missed adding `domain_data` as a parameter to `update_organization`, despite it being documented.
* We were always serializing certain parameters as `null` in the request body, even when the value is not passed.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/organizations.rb
+++ b/lib/workos/organizations.rb
@@ -74,12 +74,13 @@ module WorkOS
       # @param [Array<Hash>] domain_data List of domain hashes describing an orgnaization domain.
       # @option domain_data [String] domain The domain that belongs to the organization.
       # @option domain_data [String] state The state of the domain. "verified" or "pending"
+      # @param [String] name A unique, descriptive name for the organization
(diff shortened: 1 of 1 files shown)


lib/workos/organizations.rb
```diff
@@ -74,29 +74,37 @@ def get_organization(id:)
...
       def create_organization(
         domain_data: nil,
         domains: nil,
         name:,
         allow_profiles_outside_organization: nil,
         idempotency_key: nil
       )
-        warn_deprecation '`domains` is deprecated. Use `domain_data` instead.' if domains
+        body = { name: name }
+        body[:domain_data] = domain_data if domain_data
```
mthadley, line 92: Previously we were unconditionally passing `domain_data`, which meant it was always in the request body, even if no value was passed.

cmatheson (approved):

PR #309, opened 2024-06-24
Add "role" to organization membership and directory sync user

ameesha:
## Description
Add "role" property to both "organization_membership" (non nullable) and "directory_user" (nullable)

Also added update_organization_membership method 
and updated "created_organization_membership" method to accept role_slug

## Documentation

```
[ ] Yes
```


PaulAsjes (approved):

lib/workos/user_management.rb
```diff
@@ -841,14 +841,37 @@ def list_organization_memberships(options = {})
...
+      #
+      # @param [String] organization_membership_id The ID of the Organization Membership.
+      # @param [String] role_slug The slug of the role to grant to this membership.
+      #
+      # @return [WorkOS::OrganizationMembership]
+      def update_organization_membership(organization_membership_id:, role_slug:)
+        request = put_request(
+          path: "/user_management/organization_memberships/#{id}",
+          body: {
+            organization_membership_id: organization_membership_id,
```
georg, line 873: The API expects an `id` field in the body, not `organization_membership_id`.
Line 871 above references an `id` variable which is not defined here.

I suggest renaming the parameter from `organization_membership_id` to `id` to keep parity with the `delete_organization_membership` method.

ameesha: Sorry about that, fixed in the 5.2.1 patch
https://rubygems.org/gems/workos

PR #323, opened 2024-08-08
Add Apple to list of providers.

MikeMcQuaid:
This may be me completely understanding this code but: this looked like an omission here so I thought I'd open a PR for discussion.

## Documentation

```
[ ] Yes
```


mthadley (approved):
That was not an intentional omission. Thanks for fixing, @MikeMcQuaid!

PR #338, opened 2024-12-23
Add GET /organization/:orgId/roles support

mattgd:
## Description
Add GET /organization/:orgId/roles support.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/lib/workos/organizations.rb
+++ b/lib/workos/organizations.rb
@@ -194,7 +194,7 @@ module WorkOS
         parsed_response = JSON.parse(response.body)
 
         roles = parsed_response['data'].map do |role|
-          ::WorkOS::Role.new(role.to_json)
+          WorkOS::Role.new(role.to_json)
         end
 
         WorkOS::Types::ListStruct.new(
--- a/lib/workos/organizations.rb
+++ b/lib/workos/organizations.rb
@@ -180,6 +180,32 @@ module WorkOS
         response.is_a? Net::HTTPSuccess
       end
 
+      # Retrieve a list of roles for the given organization.
+      #
+      # @param [String] organizationId The ID of the organization to fetch roles for.
+      def list_organization_roles(organization_id:)
+        response = execute_request(
+          request: get_request(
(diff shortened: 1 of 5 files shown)


lib/workos/organizations.rb
```diff
@@ -180,6 +180,32 @@ def delete_organization(id:)
...
+          ::WorkOS::Role.new(role.to_json)
+        end
+
+        WorkOS::Types::ListStruct.new(
+          data: roles,
+          list_metadata: {
+            after: nil,
+            before: nil,
+          },
+        )
```
mattgd, line 206: This is a bit silly. There's no pagination for this API, so perhaps I should just create a new type. Let me know your thoughts.

mthadley: I suppose you could just return the `roles` array directly if you don't think it'll ever be a paginated API.

But if there's a chance it'll become paginated later, then returning this at least prevents the need to introduce a breaking change later.

lib/workos/organizations.rb
```diff
@@ -180,6 +180,32 @@ def delete_organization(id:)
...
+          request: get_request(
+            path: "/organizations/#{organization_id}/roles",
+            auth: true,
+          ),
+        )
+
+        parsed_response = JSON.parse(response.body)
+
+        roles = parsed_response['data'].map do |role|
+          ::WorkOS::Role.new(role.to_json)
```
mthadley, line 197: Is the leading `::` needed here?

```suggestion
          WorkOS::Role.new(role.to_json)
```

mthadley (approved):

lib/workos/organizations.rb
mattgd: Oops, nope. Fixing, thanks!

lib/workos/organizations.rb
mattgd: Having to introduce a breaking change later is my worry. Pagination doesn't really make sense right now from a product perspective, but I worry if this will change as we see more use cases. I'll leave as-is for now.

PR #342, opened 2025-01-13
Add caching to remote JWKS fetch

### LINKED ISSUE #334: Cache Remote JWKS Fetch
In the Ruby SDK, `WorkOS::UserManagement.load_sealed_session` seems to fetch the remote JWKS on every load, adding about 100ms to each authenticated request.

We use Rails and for every authenticated request, we call `authorize_request` which does the following:
```
session = WorkOS::UserManagement.load_sealed_session
result = session.authenticate
@user = result[:user]
```

After implementing WorkOS, we realized all of our authenticated endpoints now take 100ms longer on the backend. It took some digging through our Sentry profiles to find that `create_remote_jwk_set` was called repeatedly:
https://github.com/workos/workos-ruby/blob/8c401c2774be8fb073ba8334222ca06edd52d7a0/lib/workos/session.rb#L26

We host our services on Render.com on GCP's `oregon-1` region.

Please advise if there is a better way to authenticate the token or if improvements can be made to the SDK. Thanks!

Screenshot 2024-11-18 at 5 12 29 PM

nicknisi:
## Description

This PR adds a cache module and tests, and then uses that to cache the return value of fetching JWKS per `client_id`.

## Documentation

```
[ ] Yes
```

Fixes #334

### DIFF
--- a/lib/workos/cache.rb
+++ b/lib/workos/cache.rb
@@ -10,10 +10,10 @@ module WorkOS
 
       # Initializes a new cache entry
       # @param value [Object] The value to store in the cache
-      # @param expires_in [Integer, nil] The expiration time for the value in seconds, or nil for no expiration
-      def initialize(value, expires_in)
+      # @param expires_in_seconds [Integer, nil] The expiration time for the value in seconds, or nil for no expiration
+      def initialize(value, expires_in_seconds)
         @value = value
-        @expires_at = expires_in ? Time.now + expires_in : nil
+        @expires_at = expires_in_seconds ? Time.now + expires_in_seconds : nil
       end
 
       # Checks if the entry has expired
--- a/lib/workos/cache.rb
+++ b/lib/workos/cache.rb
@@ -8,6 +8,9 @@ module WorkOS
     class Entry
       attr_reader :value, :expires_at
 
+      # Initializes a new cache entry
+      # @param value [Object] The value to store in the cache
+      # @param expires_in [Integer, nil] The expiration time for the value in seconds, or nil for no expiration
       def initialize(value, expires_in)
         @value = value
         @expires_at = expires_in ? Time.now + expires_in : nil
--- a/lib/workos/cache.rb
+++ b/lib/workos/cache.rb
@@ -1,0 +1,63 @@
+module WorkOS
+  module Cache
+    class Entry
+      attr_reader :value, :expires_at
+
+      def initialize(value, expires_in)
+        @value = value
+        @expires_at = expires_in ? Time.now + expires_in : nil
+      end
+
+      def expired?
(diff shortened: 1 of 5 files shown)


lib/workos/cache.rb
```diff
@@ -0,0 +1,91 @@
...
+  # The Cache module provides a simple in-memory cache for storing values
+  # This module is not meant to be instantiated in a user space, and is used internally by the SDK
+  module Cache
+    # The Entry class represents a cache entry with a value and an expiration time
+    class Entry
+      attr_reader :value, :expires_at
+
+      def initialize(value, expires_in)
+        @value = value
+        @expires_at = expires_in ? Time.now + expires_in : nil
```
PaulAsjes, line 13: I'm assuming `expires_in` is in ms? If so can we be more descriptive with naming, e.g. `ms_expires_in`?

PaulAsjes: Should also add docstring comments for these parameters.

lib/workos/cache.rb
```diff
@@ -0,0 +1,91 @@
...
+        Time.now > @expires_at
+      end
+    end
+
+    class << self
+      # Fetches a value from the cache, or calls the block to fetch the value if it is not present
+      # @param key [String] The key to fetch the value for
+      # @param expires_in [Integer] The expiration time for the value in seconds
+      # @param force [Boolean] If true, the value will be fetched from the block even if it is present in the cache
+      # @param block [Proc] The block to call to fetch the value if it is not present in the cache
```
PaulAsjes, line 30: Is `block` a Ruby term? I had to read it a few times to realise it meant calling a function or assignment to set the value if it's missing from the cache.

lib/workos/session.rb
```diff
@@ -23,7 +23,9 @@ def initialize(user_management:, client_id:, session_data:, cookie_password:)
       @session_data = session_data
       @client_id = client_id
 
-      @jwks = create_remote_jwk_set(URI(@user_management.get_jwks_url(client_id)))
+      @jwks = Cache.fetch("jwks_#{client_id}", expires_in: 5 * 60) do
```
PaulAsjes, line 26: This is really nice 🚀

PaulAsjes (changes requested):
Small naming nit, otherwise good!

lib/workos/cache.rb
nicknisi: Yeah, iirc _block_ is the term for what Ruby refers to callbacks as.

lib/workos/cache.rb
nicknisi: It's actually in seconds. I added a docstring comment!

PaulAsjes (approved):

PR #384, opened 2025-07-07
Fix request body construction to only send non-nil values

nicknisi:
## Summary
Fix multiple methods in the UserManagement module to use `.compact` when building request bodies, preventing unnecessary nil values from being sent to the API.

This addresses a customer-reported issue where `update_user` was unintentionally nullifying fields that weren't explicitly provided. The fix has been extended to all similar methods for consistency.

## What was the problem?
When calling methods like `update_user` with only some parameters:
```ruby
WorkOS::UserManagement.update_user(
  id: 'user_123',
  email_verified: true
)
```

The method was sending ALL parameters (including nil ones) to the API:
```ruby
{
  email: nil,
  first_name: nil, 
  last_name: nil,
  email_verified: true,
  external_id: nil,
  # ... more nil values
}
```

This caused the API to explicitly nullify fields that users didn't intend to change.

## What's the solution?
Add `.compact` to remove nil values before sending the request:
```ruby
{
  email_verified: true  # Only non-nil values sent
}
```

## Methods Fixed
- `update_user` - Critical fix for user-reported issue
- `create_user` - Consistency and best practice
- `send_invitation` - Remove unnecessary nil invitation parameters  
- `enroll_auth_factor` - Remove unnecessary nil TOTP parameters
- `create_organization_membership` - Remove unnecessary nil role_slug
- `create_magic_auth` - Remove unnecessary nil invitation_token

## Test plan
- [x] All existing tests pass (353/353 examples)
- [x] Added specific tests to verify only non-nil values are sent
- [x] No regressions introduced
- [x] Follows existing pattern used in `authorization_url` method

## Backward compatibility
✅ **This is a PATCH-level change** - bug fixes only, no breaking changes
- Method signatures unchanged
- Behavior improved (fixes incorrect nullification)
- All existing code will work better, not break

## Pattern consistency
This change makes all methods follow the same pattern already established in the codebase at `authorization_url:121` which uses `.compact`.

### DIFF
--- a/lib/workos/user_management.rb
+++ b/lib/workos/user_management.rb
@@ -206,7 +206,7 @@ module WorkOS
             email_verified: email_verified,
             password_hash: password_hash,
             password_hash_type: password_hash_type,
-          },
+          }.compact,
           auth: true,
         )
 
@@ -643,7 +643,7 @@ module WorkOS
             body: {
               email: email,
               invitation_token: invitation_token,
-            },
+            }.compact,
             auth: true,
           ),
         )
@@ -697,7 +697,7 @@ module WorkOS
               totp_issuer: totp_issuer,
               totp_user: totp_user,
               totp_secret: totp_secret,
-            },
+            }.compact,
             auth: true,
           ),
         )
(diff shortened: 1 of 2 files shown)


ericroberts (approved):
I was curious about what if someone intentionally wanted to send a nil value, but it looks like our api for update doesn't accept any nulls

Could potentially be something in the future, but this seems good for what we have now.

PR #552, opened 2026-09-05
feat: Add optional jwt_issuer option for session token validation

m0tzy:
## Summary

Lets apps opt in to validating the `iss` claim of session access tokens. Part of the cross-SDK rollout started in workos/authkit-react-router#83 (see also workos/workos-node#1694, workos/workos-python#725); default behavior is unchanged — no issuer configured means `iss` is not checked, exactly as before.

- `WorkOS::BaseClient.new(..., jwt_issuer: nil)` / `config.jwt_issuer` — a `String` or `Array<String>` of accepted issuers, exposed as `client.jwt_issuer`. `WorkOS.client` forwards it from the global configuration.
- `SessionManager#decode_jwt` adds `iss: client.jwt_issuer, verify_iss: true` to the `JWT.decode` options only when `jwt_issuer` is non-nil. ruby-jwt's `JWT::Claims::Issuer` matches against `Array(issuers)`, so a list works natively, and `InvalidIssuerError < DecodeError` maps to the existing `INVALID_JWT` reason in `Session#authenticate`.
- An explicit empty array fails closed (every token rejected) rather than disabling the check.
- Following this SDK's convention (README: "The SDK does not read these environment variables automatically"), there is no `WORKOS_ISSUER` env lookup here; pass `ENV["WORKOS_ISSUER"]` into `config.jwt_issuer` yourself if you want that.

```ruby
WorkOS.configure { |c| c.jwt_issuer = "https://api.workos.com/user_management/#{client_id}" }
# or
WorkOS::Client.new(api_key:, client_id:, jwt_issuer: ["https://api.workos.com", "https://auth.example.com"])
```

Not touched: `lib/workos/client.rb` and `rbi/**` are oagen-generated. `Client` inherits `initialize` from the hand-maintained `BaseClient`, so no generated change is needed for the kwarg; the `.rbi` signature for `BaseClient#initialize` / `SessionManager#decode_jwt` will pick up `jwt_issuer` on the next regen (CI does not type-check the rbi files).

Tests: `standardrb` clean; `rake test` 1251 runs / 0 failures (Ruby 3.4). New cases cover unset, matching, mismatched, missing `iss`, list, empty list, and `WorkOS.configure` propagation.

Link to Devin session: https://app.devin.ai/sessions/0ee38e859a9849658a7cdb2d215d89a6
Open in Devin Desktop: https://app.devin.ai/desktop/session/0ee38e859a9849658a7cdb2d215d89a6?variant=devin
Requested by: @m0tzy

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,29 @@ Anything shorter than 32 bytes (including `nil` or `""`) raises
 `ArgumentError` as soon as you load, seal, or unseal a session — sealing or
 unsealing will not silently proceed with a weakened key.
 
+### Validating the access token issuer
+
+By default the SDK verifies the signature and expiry of session access tokens
+but does not check the `iss` claim. To also require a specific issuer (or one
+of several), set `jwt_issuer` on the client — either a single string or an
+array of accepted issuers:
+
+```ruby
+WorkOS.configure do |config|
+  config.jwt_issuer = "https://api.workos.com/user_management/#{ENV["WORKOS_CLIENT_ID"]}"
+end
+
+# or per client
+client = WorkOS::Client.new(
+  api_key: ENV.fetch("WORKOS_API_KEY"),
+  client_id: ENV["WORKOS_CLIENT_ID"],
+  jwt_issuer: ["https://api.workos.com", "https://auth.example.com"]
+)
+```
+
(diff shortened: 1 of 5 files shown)


gjtorikian (approved):
Thanks!

